PR #65, opened 2026-04-20
fix: resolve CI spec-check failure for deps spec

corvid-agent:
## Summary
- Removes `depends_on: [run]` from the deps spec frontmatter — the external `spec-sync@v4.3.1` GitHub Action can't resolve it even though the `run` spec exists
- The dependency on `run::detect_project_type` is still documented in the Dependencies section
- Fixes CI failure on main introduced by PR #62

## Test plan
- [x] `cargo run -- spec check --strict` passes locally (21 specs, 0 errors, 0 warnings)
- [ ] CI spec-check should pass

### DIFF
--- a/specs/deps/deps.spec.md
+++ b/specs/deps/deps.spec.md
@@ -7,7 +7,7 @@ files:
 
 db_tables: []
 depends_on:
-  - run
+  - specs/run
 ---
 
 # Deps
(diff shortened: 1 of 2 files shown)


corvid-agent:
Superseded by #67 — spec-sync v4.3.2 properly resolves bare module names, so we can keep `depends_on: [run]` and it works. #67 bumps the CI action and cleans up all specs.

PR #265, opened 2026-04-25
feat(release): recognize plugin.toml as a first-class version source

### LINKED ISSUE #264: fledge release should know about its own manifests (plugin.toml, template.toml, fledge.toml lane repos)
## Context

Hit while dogfooding fledge to publish [`CorvidLabs/fledge-plugin-github` v0.2.0](https://github.com/CorvidLabs/fledge-plugin-github/releases/tag/v0.2.0):

```
$ fledge release minor --push --ni
error: No version found in project files or git tags. Specify an explicit version: fledge release 1.0.0

$ fledge release 0.2.0 --push --ni
📦 0.0.0 → 0.2.0
  Tag-only release (no version files found)   ← plugin.toml was right there
  ...
```

`plugin.toml` has a `[plugin].version` field, but `fledge release` doesn't recognize it. Net effect: every fledge plugin author has to (1) hand-bump `plugin.toml`, (2) commit it, (3) then run `fledge release <explicit-version>`. Fledge's own first-class manifest is being treated as a foreign format.

## The ask

`fledge release` should be the best release tool for fledge's own ecosystem. Concretely:

### 1. Plugins — `plugin.toml`

Field already exists at `[plugin].version`. Add it to the recognized version-source list so `fledge release minor` Just Works inside a plugin repo.

### 2. Templates — `template.toml`

No version field today (only a commented `min_fledge_version`). Add `[template].version` to the manifest schema, scaffold it via `fledge templates create`, and have `fledge release` detect+bump it.

### 3. Lane repos — `fledge.toml`

Lane repos use `fledge.toml` with `[tasks]`/`[lanes.*]` and no version field. For repos published via `fledge lanes publish`, the version is meaningful (consumers `lanes import` against a tag). Add an optional `[package].version` (or similar) for lane repos and detect it.

## Why this matters

- **Dogfooding signal**: fledge plugins/templates/lanes are the obvious set of repos for `fledge release` to handle natively. Anything else would be surprising.
- **One less footgun**: today nothing prevents `plugin.toml`'s version from drifting away from the git tag, since release silently skips it.
- **Discoverability**: `fledge release minor` working out-of-the-box inside `fledge plugins create`/`fledge templates create`/`fledge lanes create` outputs is the kind of compounding ergonomic that makes the whole CLI feel coherent.

## Suggested behavior

- `fledge release <bump>` inside a recognized fledge manifest repo → detect manifest, bump the right field, include the bump in the release commit.
- Multiple manifests in one repo (rare) → bump all of them, mention each in the changelog commit body.
- `--no-bump` escape hatch for the existing tag-only behavior.
- `fledge plugins/templates/lanes validate` should warn if `version` is missing on a manifest that's expected to carry one.

Happy to take a swing at this if there's appetite — would want a quick steer on whether template/lane manifests should grow a top-level `version` or whether a separate `[package]` table is preferred.

0xLeif:
Closes #264.

## Problem

Hit while dogfooding `fledge release` to publish [`fledge-plugin-github` v0.2.0](https://github.com/CorvidLabs/fledge-plugin-github/releases/tag/v0.2.0):

\`\`\`
$ fledge release minor --push --ni
error: No version found in project files or git tags. Specify an explicit version: fledge release 1.0.0

$ fledge release 0.2.0 --push --ni
📦 0.0.0 → 0.2.0
  Tag-only release (no version files found)   ← plugin.toml was right there
\`\`\`

\`plugin.toml\` has a \`[plugin].version\` field, but \`fledge release\` didn't recognize it. Every fledge plugin author had to hand-bump it, commit, then \`fledge release <explicit-version>\`.

## Fix

- **\`detect_current_version\`** now reads \`plugin.toml\` (\`[plugin].version\`) before falling back to language-specific manifests. \`plugin.toml\` is the canonical fledge-ecosystem identity.
- **\`bump_version_files\`** writes \`[plugin].version\` and (when present) Cargo.toml / package.json in the same release commit so Rust plugins like \`fledge-plugin-metrics\` keep both in sync.
- **Section-scoped bumper**: walks the file for the \`version\` line inside the \`[plugin]\` table only — a \`version\` on a \`[[commands]]\` row (or any other table) is left alone.
- **\`--no-bump\`** flag for tag-only releases (per the issue's "escape hatch" suggestion).
- **\`fledge plugins validate\`** now flags \`plugin.version\` values that don't parse as semver.

## Scope

Templates and lanes were left out (per the user's redirect). Template versioning is init-from-snapshot, not contract-versioned; lane repos already work fine with \`lanes import owner/repo@v1.0.0\`.

## Smoke test against the exact scenario

\`\`\`
$ cd fake-plugin-repo  # plugin.toml has version = "0.1.5"
$ fledge release minor --dry-run
* Would release v0.2.0 (dry run)
  Would bump: plugin.toml
  Would update CHANGELOG.md
  Would create tag v0.2.0
\`\`\`

## Test plan

- [x] \`cargo test\`: 670 unit + 157 integration (7 new release tests)
- [x] \`cargo clippy --tests -- -D warnings\` clean
- [x] \`cargo fmt --check\` clean
- [x] \`cargo run -- spec check --strict\` (29 specs, 0/0)
- [x] Manual smoke test reproduces the issue scenario and passes
- [ ] After merge: cut a real fledge-plugin-* release with \`fledge release minor\` to confirm no further hand-bumping needed

### DIFF
--- a/specs/release/release.spec.md
+++ b/specs/release/release.spec.md
@@ -1,6 +1,6 @@
 ---
 module: release
-version: 1
+version: 2
 status: active
 files:
   - src/release.rs
@@ -17,7 +17,7 @@ depends_on:
 
 ## Purpose
 
-Provides a unified release workflow: version bumping across language ecosystems, changelog generation from conventional commits, git tagging, and optional push. Supports Rust, Node/Bun, Python, Ruby, Java (Gradle/Maven), Go, Swift, and any project with git tags.
+Provides a unified release workflow: version bumping across language ecosystems, changelog generation from conventional commits, git tagging, and optional push. Supports fledge plugins (`plugin.toml`), Rust, Node/Bun, Python, Ruby, Java (Gradle/Maven), Go, Swift, and any project with git tags.
 
 ## Public API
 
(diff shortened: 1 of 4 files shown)


corvid-agent:

corvid-agent (approved):
Clean PR — well-scoped, well-tested. Approve.

**What I like:**

- Section-scoped TOML parsing (`extract_versioned_toml_section` / `replace_versioned_toml_section`) is the right call. Avoids pulling in a full TOML library while correctly handling the `[[commands]]` version-collision scenario. The tests at L1087–1132 nail this.
- `plugin.toml` checked first in `detect_current_version` — correct priority for the fledge ecosystem.
- Dual-file sync (plugin.toml + Cargo.toml) for Rust plugins prevents the exact kind of version drift that'd bite someone on a `cargo publish` after a `fledge release`.
- `--no-bump` is a clean escape hatch. The spec documents the intent clearly.
- Spec bumped to v2 with invariants 10–11 explicitly covering the scoping and sync behavior.

**Minor observations (non-blocking):**

1. **Regex per call** — `extract_versioned_toml_section` and `replace_versioned_toml_section` compile a `Regex` on each invocation. Fine for a release tool (runs once), but if these helpers ever get reused in a hot path, consider `lazy_static!` / `std::sync::LazyLock`.

2. **`trimmed.starts_with('[')` as section detector** — would false-positive on a line like `foo = [1, 2]` at zero indent. In practice `plugin.toml` is tightly controlled so this is a non-issue, but worth a comment if the helpers get extracted to a shared module later.

3. **`--no-bump` fallback version** — when `detect_current_version` fails, the synthetic `BumpResult` shows `0.0.0 → X.Y.Z`. A `--no-bump` user probably won't be confused (they asked for tag-only), but a one-line note in the `--no-bump` help text ("version display may show 0.0.0 if no manifest is found") would preempt any bug reports.

4. **Trailing inline comment** — the version-line regex `^\s*version\s*=\s*"[^"]+"\s*$` won't match `version = "0.1.0" # comment`. Unlikely in machine-generated `plugin.toml`, but the `validate` subcommand could catch it with a warning if you ever want belt-and-suspenders.

None of these are blockers. The test matrix (7 new tests, section scoping, dual-file sync, edge cases) covers the important paths well.

Ship it 🚢

— CorvidAgent (Rook)

Issue #272, opened 2026-04-26
1.0: schema_version envelope migration for existing --json outputs (tier C, breaking)


0xLeif:
## Context

Multi-model readiness review flagged \"missing schema_version in JSON outputs\" across plugins/lanes/templates as a recurring `major` concern. The introspect command was migrated additively in PR #270 (`schema_version: 1` sits alongside existing top-level keys). For commands that currently return a top-level array (`plugins list --json`, etc.), there is no additive way to add `schema_version` — the migration is **breaking**.

Doing this **before 1.0** is the only time it's free. After 1.0 the top-level shape is frozen and these arrays can't be wrapped without a 2.0.

## Migration

Each existing `--json` output that returns a top-level array gets wrapped as `{schema_version: 1, <resource>: [...]}`:

### plugins
- [ ] `plugins list --json`: `[...]` → `{schema_version: 1, plugins: [...]}`
- [ ] `plugins audit --json`: `[...]` → `{schema_version: 1, audit: [...]}`
- [ ] `plugins search --json`: `[...]` → `{schema_version: 1, results: [...]}`
- [ ] `plugins validate --json`: existing object → add `schema_version: 1` (additive — already an object)

### lanes
- [ ] `lanes list --json`: `[...]` → `{schema_version: 1, lanes: [...]}`
- [ ] `lanes run --json`: existing object → add `schema_version: 1`
- [ ] `lanes search --json`: `[...]` → `{schema_version: 1, results: [...]}`
- [ ] `lanes validate --json`: existing object → add `schema_version: 1`

### templates
- [ ] `templates search --json`: `[...]` → `{schema_version: 1, results: [...]}`
- [ ] `templates validate --json`: existing object → add `schema_version: 1`

## Out-of-tree consumers to update in the same PR

- [ ] `AGENTS.md` — every `jq '.[]'` example needs `jq '.<resource>[]'`
- [ ] `docs/src/cli-reference.md` JSON shape examples
- [ ] Any in-tree integration test that asserts on the array shape (`tests/main.rs` plugin/lane/template JSON tests)
- [ ] `Formula/fledge.rb`, `flake.nix` — no JSON consumption, no change

## Out-of-tree consumers we cannot update

- Each `CorvidLabs/fledge-plugin-*` repo whose CI or shell helpers consume `plugins list --json`, etc. — **audit each before merge.** Bump fledge minimum-version in those repos to the version that ships this migration.
[...]

PR #270, opened 2026-04-26
chore: lock 1.0 contracts (introspect schema, plugin protocol, lanes, llm, trust)

0xLeif:
## Summary

Six concrete 1.0 readiness items, identified by an independent code-review pass. **Not** a 1.0 tag — that needs a separate big review on this branch first.

| # | Blocker | Resolution |
|---|---------|------------|
| 1 | Plugin protocol compat policy | New "Compatibility Policy" section in `specs/plugin/plugin-protocol.spec.md` (v1 → v2): `fledge-v1` is additive-only, no field removal/retyping, wire format frozen, `init` message guaranteed. Field-removal requires `fledge-v2` |
| 2 | Lanes parallel + `fail_fast` semantics | Pinned in `specs/lanes/lanes.spec.md` (v9 → v10) invariant 4: in-flight siblings always finish (no cancellation), `fail_fast` governs only the *next* step. Documents existing `std::thread::scope` behavior — no code change |
| 3 | Introspect schema versioning | `schema_version: 1` emitted at top of `--json` output, additively (sits alongside `name`/`about`/`args`/`subcommands`, not nested). Old consumers untouched. Spec v1 → v2 with explicit additive-only contract |
| 4 | AI provider commitment | New "1.0 Provider Contract" in `specs/llm/context.md`: `provider = "claude"` is `claude` CLI shell-out; `provider = "ollama"` is HTTP `/api/generate`. Adding a new transport under an existing provider name is breaking and out of scope for v1; adding a new `ProviderKind` variant remains additive |
| 5 | Trust tier completeness | Renamed `TrustTier::Community` → `TrustTier::Team` with a real construction site: `TEAM_MEMBERS = ["0xLeif"]` allowlist (case-insensitive). Drops `#[allow(dead_code)]`. Spec v1 → v2 |
| 6 | Spec drift sweep | `fledge spec check --strict` clean (29/29). Fixed 4× `laness` typos in `specs/lanes/lanes.spec.md` |

## What's *not* in this PR

- No `1.0.0` tag, no version bump, no release. That waits on the bigger review pass against this branch.
- No code change to lanes parallel cancellation — pinning the doc to match the existing behavior was the explicit choice (option A from the triage).
- No Anthropic-API-direct provider — left as a future additive `ProviderKind::Anthropic` variant per the locked contract.

## Test plan

- [x] `cargo build` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean
[...]

Issue #271, opened 2026-04-26
1.0: extend --json coverage across plugins/lanes/templates (tier B)


0xLeif:
## Context

Multi-model readiness review (qwen3-coder / gpt-oss / kimi / minimax — see `.reviews/review.html`) flagged JSON-coverage fragmentation across the three pillars as a recurring `major` concern. **Tier A** (silent-ignore on `plugins install/remove/update`) is fixed in #270 follow-up branch. **This issue tracks tier B**: the additive work to bring `--json` to every remaining subcommand in plugins/lanes/templates.

Doing this before 1.0 is purely additive — current scripts that call these commands without `--json` keep working unchanged. After 1.0 this becomes additive-only forever, but missing a `--json` flag on a frozen 1.0 surface is awkward to add later.

### plugins (2 left)
- [ ] `plugins create --json` — emit `{schema_version, action: "create", path, name, description, files_created: [...]}`
- [ ] `plugins publish --json` — emit `{schema_version, action: "publish", repo: {owner, name, url}, visibility, validated: bool}`

### lanes (4 left)
- [ ] `lanes init --json` — emit `{schema_version, action: "init", project_type, lanes_added: [...], file: \"fledge.toml\"}`
- [ ] `lanes import --json` — emit `{schema_version, action: \"import\", source, imported: [\"<lane>\", ...], skipped: [\"<lane>\", ...], file: \".fledge/lanes/<owner>-<repo>.toml\"}`
- [ ] `lanes publish --json` — emit `{schema_version, action: \"publish\", repo: {owner, name, url}, visibility, lanes_published: N}`
- [ ] `lanes create --json` — emit `{schema_version, action: \"create\", path, name, description, files_created: [...]}`

### templates (4 left)
- [ ] `templates init --json` — emit `{schema_version, action: \"init\", project: {path, name}, template: {name, source, ref?}, variables_used: {...}, files_created: [...], hooks_run: [...]}`
- [ ] `templates create --json` — emit `{schema_version, action: \"create\", path, name, description, render_patterns: [...], files_created: [...]}`
- [ ] `templates list --json` — emit `{schema_version, templates: [{name, path, source: \"builtin|local|remote\"}, ...]}`. **Currently errors with \"unexpected argument\"** — flag isn't even accepted.
- [ ] `templates publish --json` — emit `{schema_version, action: \"publish\", repo: {owner, name, url}, visibility, validated: bool, templates_published: N}`

## Per-command requirements

For each:

[...]

PR #273, opened 2026-04-26
feat(1.0): --json coverage across plugins, lanes, templates (tier A + B)

0xLeif:
Tier A + Tier B from the multi-model 1.0 readiness review. Closes the JSON-coverage fragmentation flagged by gpt-oss/kimi/minimax (see \`.reviews/review.html\`, gitignored).

## Tier A — bug fix (commit fb181d9)

\`fledge plugins\` had a global \`--json\` that \`install\`, \`remove\`, and \`update\` silently ignored. Agents passing \`--json\` got back ANSI prose. Now each emits \`{schema_version: 1, action, ...}\` envelopes.

## Tier B — additive coverage (commit 523fe33)

10 more subcommands gained \`--json\`. Notably **\`templates list --json\` previously errored with \"unexpected argument\"** — fixed.

| Pillar | Commands now honoring --json |
|---|---|
| plugins | create, publish |
| lanes | init, import, publish, create |
| templates | init, create, **list**, publish |

All emit \`{schema_version: 1, action, ...}\` (or \`{schema_version: 1, <resource>: [...]}\` for \`list\`). Pattern matches the introspect contract from #270.

## Common contract (now uniform across the three pillars)

- Stdout = JSON only in JSON mode
- Stderr = warnings/progress (not suppressed — visible in interactive contexts, ignored by agents)
- JSON mode implies non-interactive (\`force = yes = true\`) — capability/confirmation prompts can't deadlock an agent
- Failure paths still bubble via anyhow → stderr with non-zero exit code
- Cancelled-by-user paths emit \`{cancelled: true}\` for \`publish\` commands (vs. erroring)

## Spec bumps

| Spec | Old → New | What changed |
|---|---|---|
| plugin | v13 → v14 | invariant 11 enumerates create + publish |
| lanes | v10 → v11 | init/import/publish/create JSON shapes documented |
| init | v7 → v8 | new invariant 7 documents JSON envelope contract |
| create_template | v1 → v2 | new invariant 10 documents JSON envelope |
| main | v6 → v7 | templates list/publish JSON shapes (live in main.rs) |

## Tests added

\`tests/main.rs\` — 7 new integration tests covering envelope shape and exit-code contracts:
- cli_plugin_create_json_emits_envelope
- cli_lanes_init_json_emits_envelope (+ error path)
- cli_lanes_create_json_emits_envelope
- cli_templates_list_json_emits_envelope (regression — was erroring)
- cli_templates_create_json_emits_envelope
- cli_templates_init_json_emits_envelope

\`*publish\` and \`lanes import\` would need a live GitHub network + state — out of scope for the cheap integration suite.

## Out of scope (still blocking 1.0)

**Tier C (#272)** — breaking envelope migration: wrapping existing top-level arrays (\`plugins list/audit/search\`, \`lanes list/run/search\`, \`templates search\`) as \`{schema_version: 1, <resource>: [...]}\`. **Last chance to break the shape before 1.0 freezes it.** Should land before tagging 1.0 and ship with an AGENTS.md / docs sweep.

## Pre-existing flake (NOT this PR)

\`src/lanes.rs::tests\` has 7 tests that fail under default \`--test-threads=8\` because one test mutates process cwd. Reproduces on main. Not introduced by this PR; filing separately.

## Test plan

- [x] cargo build, cargo fmt --check, cargo clippy --all-targets -- -D warnings — all clean
- [x] cargo test — 679 unit + 79 integration + per-spec suites, **0 failures** (with --test-threads=1 to avoid pre-existing lanes flake)
- [x] fledge spec check --strict — 29/29 clean
- [x] Manual smoke test of every new --json path
- [ ] Reviewer to spot-check \`fledge templates list --json | jq '.schema_version'\` returns \`1\`
- [ ] Reviewer to confirm \`fledge plugins remove nonexistent --json\` exits non-zero (tier A invariant)
- [ ] After merge: re-run multi-model panel via \`.reviews/run_focused.sh <model>\` for each of qwen3-coder/gpt-oss/kimi/minimax and confirm verdict moves toward GO

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Cargo.lock
 .DS_Store
 docs/book/
 .claude/
+.reviews/
--- a/specs/create_template/create_template.spec.md
+++ b/specs/create_template/create_template.spec.md
@@ -1,6 +1,6 @@
 ---
 module: create_template
-version: 1
+version: 2
 status: active
 files:
   - src/create_template.rs
@@ -53,6 +53,7 @@ Scaffolds a new fledge template project with a valid `template.toml` manifest, e
 7. Custom prompts section is only included when user opts in
 8. Scaffolded template includes example `.tera` file demonstrating variable substitution
 9. Scaffolded template includes author-facing README with testing instructions
+10. `--json` emits a single `{schema_version: 1, action: "create", path, name, description, render_patterns, include_hooks, include_prompts, files_created}` envelope on stdout. Prose suppressed; JSON mode implies non-interactive (`yes = true`)
 
 ## Behavioral Examples
 
@@ -109,3 +110,4 @@ Scaffolds a new fledge template project with a valid `template.toml` manifest, e
 | Date | Author | Change |
 |------|--------|--------|
 | 2026-04-19 | CorvidAgent | Initial spec |
+| 2026-04-25 | 0xLeif | v2: `--json` emits structured envelope (schema_version: 1) for `templates create`; prose suppressed, JSON mode implies non-interactive |
--- a/specs/init/init.spec.md
+++ b/specs/init/init.spec.md
@@ -1,6 +1,6 @@
 ---
 module: init
-version: 7
+version: 8
 status: active
 files:
   - src/init.rs
@@ -31,7 +31,7 @@ Orchestrates project creation from a template. Resolves the template, prompts fo
 
 | Type | Description |
 |------|-------------|
-| `InitOptions` | Options for project creation: name, template, output, author, org, no_git, no_install, refresh, dry_run, yes |
+| `InitOptions` | Options for project creation: name, template, output, author, org, no_git, no_install, refresh, dry_run, yes, json |
 
 ### Traits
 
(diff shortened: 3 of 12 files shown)


corvid-agent:

corvid-agent (dismissed):
**LGTM** — clean, well-scoped fix. The mechanical `if !json` / `if json` wrapping is thorough and consistent across all three commands.

Highlights:

- `install_plugin` returning `Result<serde_json::Value>` so single-install and bulk-install share the report shape while callers build their own envelope — good separation of concerns.
- `json` implying `force` (via `force = force || is_non_interactive() || json`) to prevent capability-prompt deadlocks for agents — exactly right.
- Error paths still exit non-zero in JSON mode, confirmed by `cli_plugin_remove_json_error_path_returns_nonzero`.
- Spinners conditionally wrapped in `Option` so no ANSI leakage in JSON mode.

One minor observation (non-blocking): in `update_plugins`, when a named plugin isn't found, the `bail!()` fires without emitting any JSON on stdout. The empty-registry and empty-defaults early-exit paths *do* emit a structured envelope before returning `Ok(())`. An agent parsing stdout-only would get no JSON in the named-not-found case. Not a spec violation (stderr + non-zero is compliant), but if you want full consistency for agent consumers, that path could emit a `{schema_version: 1, action: "update", error: "..."}` envelope before bailing. Low priority — feel free to defer to #271.

Spec v13 changelog and invariant 11 rewrite look accurate. Tests cover the key contracts.

🤖 CorvidAgent (Rook) — automated review

PR #275, opened 2026-04-26
feat: extend --json coverage across plugins/lanes/templates (tier B)

corvid-agent:
## Summary

- Adds `--json` flag to 10 remaining subcommands across plugins (`create`, `publish`), lanes (`init`, `import`, `publish`, `create`), and templates (`init`, `create`, `list`, `publish`)
- Each emits a `{schema_version: 1, action, ...}` envelope on stdout; errors still exit non-zero in JSON mode
- Integration tests for every new `--json` path asserting envelope shape + non-zero exit on error
- Spec updates with new invariants documenting JSON shapes for all 5 affected modules

## Test plan

- [x] `cargo test` — all 822 tests pass (679 unit + 143 integration across 8 test binaries)
- [x] `cargo clippy -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] New integration tests cover:
  - `cli_plugin_create_json_emits_envelope` — asserts schema_version, action, name, path, files_created
  - `cli_plugin_publish_json_error_path_returns_nonzero` — empty dir → non-zero exit
  - `cli_lane_init_json_emits_envelope` — asserts schema_version, action, project_type, lanes_added
  - `cli_lane_import_json_error_path_returns_nonzero` — nonexistent repo → non-zero exit
  - `cli_lane_publish_json_error_path_returns_nonzero` — empty dir → non-zero exit
  - `cli_lane_create_json_emits_envelope` — asserts schema_version, action, name, path, files_created
  - `cli_templates_list_json_emits_envelope` — asserts schema_version, templates array with name+source
  - `cli_templates_init_json_emits_envelope` — asserts schema_version, action, project, template, files_created
  - `cli_templates_init_json_error_path_returns_nonzero` — nonexistent template → non-zero exit
  - `cli_templates_create_json_emits_envelope` — asserts schema_version, action, name, path, files_created
  - `cli_templates_publish_json_error_path_returns_nonzero` — empty dir → non-zero exit

Closes #271

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #276, opened 2026-04-26
fix(json): tighten lanes import + templates list envelopes

0xLeif:
## Summary

Two follow-ups from the multi-model review on #275 (which is otherwise redundant with already-merged #273 — see closure note there).

- **`lanes import --json`**: always emit a string `file` path (the computed `.fledge/lanes/<safe>.toml` target) instead of `null` when every lane was skipped. New `written: bool` field distinguishes the no-op case. The path is a pure function of source owner/repo/subpath, so `null` was misleading — the file is *where* this import targets, not *what* was written.
- **`templates list`**: empty case now exits 0 in both modes. JSON returns `{schema_version: 1, templates: [], hint}`; CLI prints a friendly "No templates configured" message + the same hint. Previously both modes bailed non-zero, forcing agents to special-case a list query into an error path.

## Spec bumps

- `lanes` v13 → v14: documents the `file` (always-string) + `written: bool` invariant
- `main` v8 → v9: documents the `templates list` empty-list semantics

## Test plan

- [x] `cargo build` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo test --test main` — 79/79 pass
- [x] `cargo test --test lanes` — 8/8 pass
- [x] `cargo test --test templates` — 17/17 pass
- [x] `cargo test --bin fledge -- lanes::` — 57/57 pass
- [x] `fledge spec check --strict` — 29/29 clean
- [x] Manual smoke: `fledge templates list --json` returns `schema_version: 1` envelope with builtin templates

## Why not part of #275?

PR #275 conflicts with `main` (mergeStateStatus: DIRTY) because the same tier-B JSON coverage shipped in #273. Resolving 25 files of overlap to land these two small fixes would be churn. Direct fix on `main` is cleaner; #275 will be closed.

### DIFF
--- a/specs/lanes/lanes.spec.md
+++ b/specs/lanes/lanes.spec.md
@@ -1,6 +1,6 @@
 ---
 module: lanes
-version: 13
+version: 14
 status: active
 files:
   - src/lanes.rs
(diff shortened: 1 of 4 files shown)


corvid-agent:

corvid-agent (approved):
**LGTM — clean, minimal, correct.**

Reviewed the diff for both changes:

### `lanes import --json` (`src/lanes.rs`)
- Hoisting the `safe_name` / `relative_file` computation above the empty-check is sound — it's a pure function of `owner`, `repo`, and `subpath`, all available at that point. No side effects moved.
- The `"file": relative_file` + `"written": false` contract is cleaner than `null` — agents no longer need to guess the target path when every lane was skipped.
- Success path correctly adds `"written": true`.

### `templates list` (`src/main.rs`)
- Replacing `anyhow::bail!` with exit-0 + hint is the right call — a list returning empty isn't an error, and forcing non-zero made agents treat a normal query as a failure.
- The `hint` field being conditional (only present when `available.is_empty()`) is a reasonable choice — keeps the non-empty envelope slim while still guiding agents in the empty case.
- `let mut result` + index assignment for the optional hint field is idiomatic serde_json.

### Specs
- Both bumps (lanes v13→v14, main v8→v9) have clear changelog entries documenting the invariant changes. Good.

No issues found. All test suites green per the PR description. Ship it.

PR #274, opened 2026-04-26
feat(1.0)!: wrap --json outputs in schema_version envelope (tier C)

0xLeif:
**BREAKING.** Closes #272 (tier C of the multi-model 1.0 readiness review). Stacks on top of PR #273 (tier A + B).

## What changes for consumers

Six commands previously returned a top-level JSON array. They now return a JSON object with the array under a named key. Update `jq` paths in any scripts:

| Command | Before | After |
|---|---|---|
| `fledge plugins list --json` | `[…]` | `{schema_version:1, plugins:[…]}` |
| `fledge plugins audit --json` | `[…]` | `{schema_version:1, audit:[…]}` |
| `fledge plugins search --json` | `[…]` | `{schema_version:1, results:[…]}` |
| `fledge lanes list --json` | `[…]` | `{schema_version:1, lanes:[…]}` |
| `fledge lanes search --json` | `[…]` | `{schema_version:1, results:[…]}` |
| `fledge templates search --json` | `[…]` | `{schema_version:1, results:[…]}` |
| `fledge templates validate --json` | `[{report}, …]` | `{schema_version:1, reports:[…]}` |

Migration: `jq '.[]'` → `jq '.<resource>[]'`.

Additive (no script breakage) — `schema_version: 1` added to outputs that were already objects:

- `fledge plugins validate --json`
- `fledge lanes run --json`
- `fledge lanes validate --json`

## Why now (the 1.0 case)

`schema_version` cannot be added to a top-level array additively — there's no key to put it next to. Doing this in 0.x is the **last** time it's free. Once 1.0 ships, the top-level shape is frozen and a future migration would require a major version bump.

Within `schema_version: 1`, new fields are additive — consumers must ignore unknown keys. Removing or retyping a field bumps `schema_version` to `2`.

## Audit of out-of-tree consumers

`gh search code` against `org:CorvidLabs` for `fledge plugins list --json | jq` and similar patterns: **zero hits in external plugin repos.** The only in-tree consumer of the bare-array shape was a `jq '.[].trust_tier'` example in `specs/trust/testing.md` — updated in this commit.

## Files touched

- **Code**: `src/{plugin,lanes,validate,main}.rs`
- **Specs (bumped)**: `plugin v14→v15`, `lanes v11→v12`, `main v7→v8`, `validate v1→v2`
- **Docs**: `AGENTS.md` and `docs/src/agents.md` JSON-shape tables updated; `CHANGELOG.md` migration table added
[...]

PR #277, opened 2026-04-26
fix(lanes): honor --json on dry-run path (1.0 contract)

0xLeif:
## Summary

`fledge lanes run <name> --json --dry-run` previously ignored `--json` and emitted human-readable prose, breaking the contract that `--json` always means parseable stdout. Surfaced by the multi-model 1.0 readiness review (gpt-oss flagged it as the only inconsistency in the lanes JSON envelope shapes).

### Before
\`\`\`
$ fledge lanes run ci --json --dry-run
Lane: ci — CI pipeline
  1. build (task)
\`\`\`

### After
\`\`\`json
{
  "schema_version": 1,
  "lane": "ci",
  "description": "CI pipeline",
  "total_steps": 1,
  "fail_fast": true,
  "dry_run": true,
  "steps": [
    { "step": 1, "kind": "task", "name": "build" }
  ]
}
\`\`\`

Per-step shape: \`{step, kind: "task"|"inline"|"parallel", name, items?}\`. \`duration_ms\` is omitted in dry-run mode (nothing executed). Real-run shape unchanged; spec example tightened to show the full per-step record (`{step, name, success, duration_ms, error}`).

## Spec bumps

- \`lanes\` v13 → v15. **Stacks on #276** (which bumps to v14). Merge order should be #276 → this PR. If #276 lands second, this PR's spec changelog needs a one-line renumber.

## What was *not* fixed

The multi-model review surfaced a few other concerns; investigation showed most were stale or hallucinated:

| Concern | Reality |
|---|---|
| \`min_fledge_version\` parsed but not enforced | **Already enforced** at \`src/init.rs:542\` via \`check_fledge_version()\` |
| \`files.copy\` field unused | **Field doesn't exist** — \`FileRules\` only has \`render\` and \`ignore\` |
| Per-step \`duration_ms\` not specified | **Already emitted**; spec example just hid it behind \`steps: [...]\` — now shown |
| Plugin/lane envelope inconsistencies | **Fixed** in tier C (#274) |

## Test plan

- [x] \`cargo build\` — clean
- [x] \`cargo fmt --check\` — clean
- [x] \`cargo clippy --all-targets -- -D warnings\` — clean
- [x] \`cargo test\` — 848/848 pass (679 unit + 169 integration; 1 ignored, pre-existing)
- [x] \`fledge spec check --strict\` — 29/29 clean
- [x] New test \`cli_lane_run_json_dry_run_emits_envelope\` — covers task/parallel/inline step kinds
- [x] Manual smoke: confirmed JSON output parses with \`jq\`

### DIFF
--- a/specs/lanes/lanes.spec.md
+++ b/specs/lanes/lanes.spec.md
@@ -1,6 +1,6 @@
 ---
 module: lanes
-version: 14
+version: 15
 status: active
 files:
   - src/lanes.rs
@@ -188,9 +188,13 @@ $ fledge lanes publish
 ✅ . — valid (2 lanes)
 ➡️ Publishing 2 lanes as owner/my-lanes
 
-# Run a lane with JSON output
+# Run a lane with JSON output (each step record has step/name/success/duration_ms/error)
 $ fledge lanes run ci --json
-{"lane": "ci", "description": "Full CI pipeline", "total_steps": 3, "success": true, "duration_ms": 4733, "fail_fast": true, "steps": [...], "failures": []}
+{"schema_version": 1, "lane": "ci", "description": "Full CI pipeline", "total_steps": 3, "success": true, "duration_ms": 4733, "fail_fast": true, "steps": [{"step": 1, "name": "lint", "success": true, "duration_ms": 245, "error": null}, ...], "failures": []}
+
+# Dry-run with JSON (plan only — no duration_ms, includes dry_run: true)
+$ fledge lanes run ci --json --dry-run
+{"schema_version": 1, "lane": "ci", "description": "Full CI pipeline", "total_steps": 3, "fail_fast": true, "dry_run": true, "steps": [{"step": 1, "kind": "task", "name": "lint"}, {"step": 2, "kind": "parallel", "items": [{"kind": "task", "name": "fmt"}]}, ...]}
 ```
 
 ## Error Cases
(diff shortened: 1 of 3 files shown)


corvid-agent:

corvid-agent (dismissed):
Clean fix, well-scoped. Approve.

**Code review notes:**

1. **`to_string_pretty` vs compact** — The dry-run path uses `serde_json::to_string_pretty` (multi-line indented output). Worth verifying this matches the real-run path in `execute_lane` for consistency. Both are valid JSON and parse fine with `jq`, but consumers diffing dry-run vs real-run output might not expect the formatting difference. Minor nit — not blocking.

2. **`description` empty-string vs sentinel** — When `description` is `None`, the JSON path emits `""` while the prose path emits `"(no description)"`. The empty string is the right call for machine-readable output, so this is correct. Just noting it's intentional.

3. **Step shape is solid** — `{step, kind, name}` for task/inline, `{step, kind, items}` for parallel, no `duration_ms` in dry-run. All consistent with the spec and the real-run envelope shape (minus execution fields). 

4. **Test coverage is good** — `cli_lane_run_json_dry_run_emits_envelope` exercises all three step kinds (task, parallel, inline), validates the envelope structure, and checks `duration_ms` is absent. The assertion `steps[0]["duration_ms"].is_null()` correctly confirms omission (serde_json returns `Value::Null` for missing keys).

5. **Spec changelog** — v14/v15 stacking with #276 is documented clearly. Merge order dependency (#276 first) is called out in the PR description.

No issues found. Ship it.

PR #280, opened 2026-04-26
docs: sync README/AGENTS/docs to current state, drop em-dashes

0xLeif:
User-facing docs only. No code changes.

## What changed

**1. Content sync.** README and the docs/ book had stale envelope shapes and out-of-date claims (e.g. "every read command exposes --json" when after tier-D every command does). Updated to match current main:

- README quick-start, six-pillars table, plugin list
- AGENTS.md was already current after #279
- docs/src/agents.md envelope table now matches AGENTS.md
- docs/src/doctor.md JSON example shows the post-tier-D envelope
- docs/src/introduction.md and pillars.md tightened

**2. Voice.** Rewrote the high-impact files (README, AGENTS, intro, pillars, agents, doctor) to drop marketing-flavored phrasing and read as direct technical prose.

**3. Em-dash sweep across all user-facing markdown.** Replaced the dash-with-spaces pattern with ", " or ": " (for term/description bullets) or restructured into separate sentences. README, AGENTS, and 14 docs/src/ files now contain zero em-dashes.

## Out of scope

Specs (`specs/*/*.md`) still contain em-dashes, mostly in historical change-log entries. Not touched here. Opening a separate sweep PR if we want those too.

## Test plan

- [x] cargo test, 851 still pass
- [x] fledge spec check --strict, 29/29 clean
- [x] grep for em-dashes in user-facing markdown returns zero
- [x] No code files modified

PR #281, opened 2026-04-26
feat(release): add --json flag

0xLeif:
Closes the last known additive gap in the agent surface. Both the dry-run and real-run paths now emit a structured envelope when \`--json\` is passed.

## What changed

\`release\` gained \`--json\`:

**Dry-run envelope:**
\`\`\`json
{
  "schema_version": 1,
  "action": "release",
  "dry_run": true,
  "version": "0.2.0",
  "no_bump": false,
  "files_to_bump": ["Cargo.toml"],
  "will_changelog": true,
  "will_tag": true,
  "will_push": false,
  "tag": "v0.2.0"
}
\`\`\`

**Real-run envelope:**
\`\`\`json
{
  "schema_version": 1,
  "action": "release",
  "dry_run": false,
  "version": "0.2.0",
  "old_version": "0.1.0",
  "files_bumped": ["Cargo.toml"],
  "changelog_updated": true,
  "commit_created": true,
  "tag_created": true,
  "tag": "v0.2.0",
  "pushed": false
}
\`\`\`

Helper functions (\`generate_changelog_entry\`, \`create_release_commit\`, \`create_tag\`, \`push_release\`) gained a \`quiet\` param threaded from \`opts.json\` so the JSON envelope is the only thing on stdout in \`--json\` mode.

## Spec

\`release\` v2 to v3.

## Test plan

- [x] \`cargo build\` clean
- [x] \`cargo fmt --check\` clean
- [x] \`cargo clippy --all-targets -- -D warnings\` clean
- [x] \`cargo test\` 853 pass (+2 new: \`cli_release_dry_run_json_emits_envelope\`, \`cli_release_dry_run_json_no_bump_flag\`)
- [x] \`fledge spec check --strict\` 29/29 clean
- [x] Manual smoke: \`fledge release 0.2.0 --dry-run --json | jq\` parses cleanly with all expected fields

## Note on docs

Touched \`docs/src/ship.md\` and \`docs/src/cli-reference.md\` to add the \`--json\` documentation. Pre-existing em-dashes in those files are not cleaned in this PR. PR #280 (the doc-sync PR) handles that. If #280 lands first this PR will rebase cleanly.

### DIFF
--- a/docs/src/cli-reference.md
+++ b/docs/src/cli-reference.md
@@ -579,22 +579,28 @@ fledge changelog --tag v0.7.0
 
 ### fledge release `<bump>`
 
-Cut a release, bump version, generate changelog, create a git tag, and optionally push.
+Cut a release. Bump version, generate changelog, create a git tag, and optionally push.
 
 ```
 fledge release <bump> [OPTIONS]
 ```
 
 **Arguments:**
-- `<bump>` - Version bump: `major`, `minor`, `patch`, or an explicit version (e.g. `1.0.0`)
+- `<bump>`: Version bump: `major`, `minor`, `patch`, or an explicit version (e.g. `1.0.0`)
 
 **Options:**
-- `--dry-run` - Show what would happen without making changes
-- `--no-tag` - Skip creating a git tag
-- `--no-changelog` - Skip changelog generation
-- `--push` - Push commit and tag to remote after release
-- `--pre-lane <NAME>` - Run a lane before releasing (e.g. `ci`)
-- `--allow-dirty` - Allow releasing with uncommitted changes
+- `--dry-run`: Show what would happen without making changes
+- `--no-tag`: Skip creating a git tag
+- `--no-changelog`: Skip changelog generation
+- `--no-bump`: Skip bumping any version files (tag-only release)
+- `--push`: Push commit and tag to remote after release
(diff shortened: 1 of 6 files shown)


corvid-agent:

corvid-agent (dismissed):
Clean PR — good envelope design, tests, and docs. A few observations:

**Looks good:**
- Hoisting `detect_version_files` before the dry-run/JSON branch is a nice refactor — both paths need it now and the nesting is reduced.
- `generate_changelog_entry` returning `Result<bool>` is the right call for feeding `changelog_updated` into the envelope without tracking separate state.
- `quiet` threading through helpers is straightforward. No behavioral changes when `json: false`.
- Both integration tests are solid — they validate the full envelope structure, not just "it didn't crash."

**Minor observations (non-blocking):**
- The dry-run envelope includes `no_bump` but the real-run envelope doesn't. Consumers can infer it from `files_bumped: []`, so it's fine, but worth noting the asymmetry in case agents rely on a consistent field set across both modes.
- `commit_created` is hardcoded to `true` in the real-run envelope. If we ever reach that line the commit did succeed (errors propagate), so this is correct — just flagging it's not dynamically tracked.
- Tests only cover the dry-run JSON path. Real-run JSON would need git state teardown so this is reasonable, but a future test for the real-run envelope shape (even in a temp repo) would close the gap.

LGTM. Ship it. 🚢

PR #283, opened 2026-04-26
fix: propagate --json to pre-release lane, document release JSON shapes

corvid-agent:
## Summary
- **release.rs**: `run_pre_lane` now accepts and propagates the `--json` flag — its pretty-print output no longer contaminates the JSON envelope
- **AGENTS.md**: removed `release` from the no-JSON list, added `release --json` / `release --dry-run --json` / `lanes run --dry-run --json` shape rows, added the "errors are always plain stderr" rule

Follow-up to #282 — items 1–4 from the ship checklist.

## Test plan
- [x] All 854 tests pass
- [x] Clippy clean, fmt clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #284, opened 2026-04-27
fix(release): silence pre-release lane on --json (single-envelope contract)

0xLeif:
## Summary
- `fledge release --json --pre-lane <lane>` no longer emits a second JSON envelope on stdout. The release envelope is the contract; the lane is an internal gate.
- New `crate::lanes::run_for_pre_release(name, dry_run)` runs steps with subprocess output suppressed and emits no envelope of its own. Failure bails with plain stderr per the envelope rules.
- Adds the missing `templates init` / `templates create` / `templates publish` rows to the AGENTS.md core-commands table.
- Tightens the error-output rule: `lanes run --json` does emit a final envelope on failure, so "exit code is the contract, not stdout presence."
- New regression tests: `cli_release_dry_run_json_with_pre_lane_emits_single_envelope`, `cli_release_json_with_failing_pre_lane_bails_with_plain_stderr`.

Follow-up to #283 — review found the lane envelope leaked into stdout when combined with `--json`, breaking single-object JSON parsers.

## Test plan
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo test` — all 856 tests pass
- [x] `fledge spec check` — 29 specs, 0 errors
- [x] `fledge lanes run pre-commit` passes
- [x] Manual: `fledge release --dry-run --json --pre-lane smoke` emits one envelope; failing lane produces empty stdout + plain stderr + non-zero exit

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,6 +77,9 @@ Specs (`specs/<name>/*.spec.md` and companion files) are the source of truth for
 | `fledge templates list --json` | `{schema_version: 1, templates: [{name, description, source, source_ref, path}]}` | Listing available templates |
 | `fledge templates search --json` | `{schema_version: 1, results: [...]}` (same shape as plugins search) | GitHub search for `fledge-template`-tagged repos |
 | `fledge templates validate --json` | `{schema_version: 1, reports: [{path, template, errors, warnings}]}` | CI gate before publish |
+| `fledge templates init <template> --json` | `{schema_version: 1, action: "init", project: {name, path}, template: {name, source, version}, variables_used, files_created, git_initialized, hooks_run}` | Scaffolding a new project |
+| `fledge templates create --json` | `{schema_version: 1, action: "create", path, name, description, render_patterns, include_hooks, include_prompts, files_created}` | Creating a new template skeleton |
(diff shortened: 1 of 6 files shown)


corvid-agent:

corvid-agent (dismissed):
Clean fix, approving.

**What I checked:**

- `run_for_pre_release` correctly validates the lane, short-circuits on dry_run (release handles its own dry-run envelope), and delegates to `execute_lane_silent` for real runs.
- `execute_lane_silent` mirrors the step iteration and `fail_fast` semantics of `execute_lane_json` without emitting anything to stdout — the contract is clear.
- `run_pre_lane` in `release.rs` branches early on `json`, so the non-json path no longer needs `if !json` guards. The `json: false` on the fallback `LaneAction::Run` is correct since we already returned for `json == true`.
- Both regression tests are well-structured: the single-envelope test parses stdout as one JSON value (would panic on a double-envelope), and the failure test asserts empty stdout + non-zero exit + lane name in stderr.
- AGENTS.md error-output rule update is a good catch — "exit code is the contract" is the right framing now that `lanes run --json` does emit on failure.
- Spec bumped to v4 with clear changelog entry.

**Minor observation (not blocking):** `execute_lane_silent` and `execute_lane_json` share ~80% structure. If more execution modes appear later, extracting a shared skeleton that takes an output strategy might be worth it. Fine as-is for two variants.

LGTM 🦅

PR #285, opened 2026-04-27
fix(1.0)!: finalize envelope shapes before tagging

0xLeif:
## Summary

Three breaking envelope changes discovered in the pre-1.0 audit. PR A of three (B = doc-only AGENTS.md additions, C = tag 1.0).

### 1. Cancelled-path uniformity
`plugins publish --json` and `templates publish --json` previously emitted **different key sets** depending on whether the user confirmed or cancelled. Cancelled dropped `plugin`/`template`, `topic`, `install_hint`/`use_hint`, and replaced `created`/`private` with `exists`. Consumers couldn't read the same keys across outcomes.

Now both paths emit the full key set; `cancelled: bool` is the only discriminator.

```jsonc
// success and cancelled now share these keys
{
  "schema_version": 1,
  "action": "publish",
  "cancelled": false | true,
  "repo": {"owner", "name", "url", "created", "private"},
  "plugin" | "template": {...},
  "topic": "...",
  "install_hint" | "use_hint": "..."
}
```

### 2. ` tier` → `trust_tier`
`plugins install --json` (single + defaults) and `lanes import --json` were the holdouts. Every other plugin/lane envelope (`list`, `audit`, `search`) uses `trust_tier`. Standardize.

### 3. `lanes list --json steps` → `step_count`
The bare key `steps` read like an array but emitted `lane.steps.len()`. Full step detail lives in `lanes run --dry-run --json`. Renamed for clarity.

## Spec bumps

- `lanes` v17
- `plugin` v16
- `templates` v6

AGENTS.md updated for the changed shapes (`lanes list`, `templates publish`). The remaining doc gaps from the audit (missing rows for `lanes init/import/publish/create`, `plugins install/update/remove/publish/create`, plus `ai status` `*_source` clarification, `introspect` shape, `review` error case, `work start issue?`) are deferred to **PR B**.

## Test plan
- [x] `cargo build` clean
- [x] `fledge lanes run pre-commit` (fmt + clippy + test + spec-check) green locally
- [x] `specsync check --force --strict` reports 0 warnings
- [ ] CI green
- [ ] Post-merge: tag v1.0.0

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,7 +70,7 @@ Specs (`specs/<name>/*.spec.md` and companion files) are the source of truth for
 | `fledge plugins audit --json` | `{schema_version: 1, audit: [{name, version, trust_tier, capabilities, has_lifecycle_hooks, ...}]}` | Capability/hook audit |
 | `fledge plugins search --json` | `{schema_version: 1, results: [{name, full_name, stars, trust_tier, ...}]}` | GitHub search for `fledge-plugin`-tagged repos |
 | `fledge plugins validate --json` | `{schema_version: 1, path, plugin_name, errors, warnings}` | CI gate before publish |
-| `fledge lanes list --json` | `{schema_version: 1, lanes: [{name, description, steps, fail_fast, source?, trust_tier}]}` | Discovering lanes available to run |
+| `fledge lanes list --json` | `{schema_version: 1, lanes: [{name, description, step_count, fail_fast, source?, trust_tier}]}`. `step_count` is an integer; full step detail lives in `lanes run --dry-run --json` | Discovering lanes available to run |
 | `fledge lanes search --json` | `{schema_version: 1, results: [...]}` (same shape as plugins search) | GitHub search for `fledge-lane`-tagged repos |
 | `fledge lanes run <name> --json` | `{schema_version: 1, lane, success, duration_ms, fail_fast, steps: [...], failures: [...]}` | Running the project's own CI pipeline |
 | `fledge lanes validate --json` | `{schema_version: 1, path, lane_count, errors, warnings}` | CI gate before publish |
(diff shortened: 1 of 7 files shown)


corvid-agent:

corvid-agent (approved):
Clean, well-scoped set of breaking changes for the 1.0 contract freeze. Reviewed the diff — three items look correct:

**1. `steps` → `step_count` (`lanes list --json`)** — Single-site rename in `lanes.rs`. Removes the ambiguity where an integer count looked like it should be an array. Good call.

**2. `tier` → `trust_tier`** — Both branches of `lanes import --json` updated consistently (all-skipped path and success path). `plugins install --json` also updated. Aligns with every other plugin/lane envelope.

**3. Cancelled-path uniformity** — Both `plugins publish` and `templates publish` cancelled paths now emit the full key set (`repo`, `plugin`/`template`, `topic`, `install_hint`/`use_hint`). `exists: true` correctly replaced by `created: false` + `private`. The success paths get `cancelled: false` added. Consumers can now read the same keys regardless of outcome — big improvement.

**One thing to double-check:** The diff shows `"cancelled": false` being *added* to the success path for `templates publish` (main.rs:1736), but the cancelled path's `"cancelled": true` is outside the diff hunk context. For plugins publish this is fine (added in v14), but I don't see a prior templates spec version that added `cancelled: true` to the cancelled path. Worth confirming it's already present in the full source — if not, the cancelled envelope would be missing its discriminator key.

**AGENTS.md:** `lanes list` and `templates publish` rows updated correctly. The remaining doc gaps (`plugins install/publish`, `lanes import`, etc.) are explicitly deferred to PR B — just flagging so those don't slip.

Specs bumped cleanly (lanes v17, plugin v16, templates v6) with thorough changelog entries.

LGTM — approve with the minor verification note above.

🤖 Reviewed by CorvidAgent (Opus 4.6)

PR #294, opened 2026-04-28
fix(spec): resolve nested specs + organize change logs (closes #291)

### LINKED ISSUE #291: fledge review fails with nested spec paths (plugin-protocol)
## Problem
`fledge review --base main` fails when spec auto-detection encounters nested spec paths like `specs/plugin-protocol/`. The spec resolver looks for `specs/plugin-protocol/` directory instead of the correct `specs/plugin/plugin-protocol.spec.md`.

## Expected behavior
Spec auto-detection should correctly resolve spec files in nested directory structures, e.g., recognizing that `specs/plugin-protocol/` should map to `specs/plugin/plugin-protocol.spec.md`.

## Steps to reproduce
1. Have a spec structure with nested paths like `specs/plugin/plugin-protocol.spec.md`
2. Run `fledge review --base main` in a repo with this structure
3. Observe spec resolution failure on the nested path

## Environment
- fledge version: 1.0.x
- Spec structure: `specs/plugin/plugin-protocol.spec.md` (and similar nested specs)

0xLeif:
## Summary
- Fix `fledge review` spec auto-detection on nested specs like `specs/plugin/plugin-protocol.spec.md` (closes #291). `IndexEntry` now carries each spec's on-disk path; `specs_for_changed_files` and `load_module_bundle` resolve via that path instead of assuming `module name == directory name`.
- Standardize every spec's Change Log to the canonical `| Version | Date | Changes |` table, sorted newest-first, with em-dashes replaced by commas per the project writing style.

## Test plan
- [x] `cargo test` (868 pass)
- [x] `cargo clippy -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo run -- spec check --strict` (29 specs, 0 errors, 0 warnings)
- [x] Three regression tests added covering nested-spec resolution by `files:`, shared-dir match for both specs, and `load_module_bundle` for the nested name
- [x] Verified live: `cargo run -- spec list` shows both `plugin` and `plugin-protocol` entries with correct paths


corvid-agent:

corvid-agent (approved):
**Reviewed by CorvidAgent (Team Alpha Lead)**

Clean fix, well-tested. Approving.

## The core fix

The root cause was that `specs_for_changed_files` and `load_module_bundle` assumed `module name == directory name`, constructing paths like `<specs_dir>/<name>/`. This breaks for sub-specs like `specs/plugin/plugin-protocol.spec.md` where `plugin-protocol` has no matching directory.

The fix stores each spec's actual on-disk `path` in `IndexEntry` and resolves from that — correct approach. Making the data model self-describing instead of relying on naming conventions is the right design.

## What I checked

- **`specs_for_changed_files`**: Now uses `entry.path.parent()` + `strip_prefix(root)` to derive the directory prefix. Handles Windows paths via `replace('\\', "/")`. The `unwrap_or(false)` fallback is safe.
- **`load_module_bundle`**: Now resolves through `collect_index()` instead of guessing the path. Falls back to the conventional location for error messages when not found — good UX preservation.
- **Three regression tests** cover the exact issue #291 scenarios: nested spec resolution by `files:`, shared-dir matching for both specs, and `load_module_bundle` for the nested name. All well-structured.
- **Spec updates**: `spec.spec.md` bumped to v8 with new invariant 15, `review.spec.md` bumped to v10 — both accurately describe the new behavior.
- **Change log standardization**: Em-dashes → commas, canonical `| Version | Date | Changes |` format, newest-first sort. Consistent across all 20+ spec files.

## Minor observations (non-blocking)

1. `load_module_bundle` now calls `collect_index()` on every invocation (full filesystem scan) where before it did a direct path lookup. Fine for current scale, but worth noting if spec count grows significantly.
2. When two specs share a directory, companion files (like `requirements.md`) are included in both bundles — this is the correct behavior given the shared directory layout, just worth documenting awareness of.

868 tests passing, clippy clean, `spec check --strict` green. Ship it. 🚀

PR #296, opened 2026-04-28
fix(plugin): pre-1.0 security hardening

corvid-agent:
## Summary

Three security fixes that should land before 1.0:

- **Remove `--json` from auto-grant** — `install_plugin` was lumping `--json` mode with `--force`, silently granting all capabilities to agents. Now `--json` only controls output format.
- **Pin default plugins** — `fledge-plugin-github` pinned to `v0.2.1`. Added doc comment requiring pinning once a release exists (deps/metrics still track HEAD until they get their first release).
- **Capability re-prompt on update** — `fledge plugins update` now detects when a pulled version requests capabilities the user never granted and prompts for approval. Non-interactive mode fails safely instead of silently inheriting old grants. Updated capabilities are persisted to the registry.

Also fixes the `is_default` matcher and tests to handle `@ref` suffixes in `DEFAULT_PLUGINS` entries.

## Test plan

- [x] All 693 existing tests pass
- [x] Clippy clean, fmt clean
- [x] `default_plugins_are_well_formed` test updated to allow pinned refs
- [x] `default_source_match_recognizes_shorthand` test updated with ref-aware matcher

### DIFF
--- a/src/plugin.rs
+++ b/src/plugin.rs
@@ -14,12 +14,12 @@ use crate::trust::{determine_trust_tier, parse_source_ref, TrustTier};
 /// installs all of them so a fresh fledge install gets back to feature
 /// parity with v0.14 in one command.
 ///
-/// Pin every entry to a release tag (`owner/repo@vX.Y.Z`) once one
-/// exists — unpinned entries track HEAD and are a supply-chain risk.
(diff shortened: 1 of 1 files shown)


corvid-agent:

corvid-agent:
CI all green — thanks Corvin! Ready for human review.

PR #307, opened 2026-04-29
fix: install script respects detected architecture

corvid-agent:
## Summary
- `artifact_name()` now uses the detected `$arch` for Linux and Windows instead of hardcoding `x86_64`
- Adds explicit detection of 32-bit systems (`i386`/`i686`) with a clear error message instead of the generic "unsupported architecture"
- When aarch64 Linux or Windows binaries are published in the future, the script will just work

## Test plan
- [ ] Run `install.sh` on x86_64 Linux — should download `fledge-linux-x86_64`
- [ ] Run `install.sh` on aarch64 Linux — should attempt `fledge-linux-aarch64` (will fail until we publish that binary)
- [ ] Run `install.sh` on a 32-bit system — should show "32-bit systems are not supported"

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -67,16 +67,17 @@ detect_arch() {
     case "$uname_m" in
         x86_64|amd64)  echo "x86_64" ;;
         aarch64|arm64) echo "aarch64" ;;
-        *) err "unsupported architecture: $uname_m" ;;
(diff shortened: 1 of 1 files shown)


corvid-agent:

corvid-agent:
All CI green — thanks Corvin! Merging this one.

PR #322, opened 2026-05-01
fix: post-review cleanup (clippy, flaky tests, lanes import, timeouts)

0xLeif:
## Summary

A handful of fixes surfaced by a full review of the project. None change happy-path behavior — together they close a CI gap, a flaky test, an auth-bypass on community lane import, and several unbounded network calls.

- **Clippy gap.** `fledge.toml`'s `lint = "cargo clippy -- -D warnings"` skipped test code. Bumped to `--all-targets`, fixed the `module_inception` it surfaces in `src/release/tests.rs` (drops the redundant inner `mod tests {}` so the path is `release::tests::*` instead of `release::tests::tests::*`).
- **Flaky `release::tests::full_release_flow`.** The cwd `Mutex` lived inside `release/tests.rs`, but `lanes/tests.rs` mutates `std::env::current_dir` without it — they raced on parallel cargo runs. Hoisted into a new `src/test_support.rs` so both modules share a lock; wrapped `with_cwd`'s closure in `catch_unwind` so a panicking test still restores cwd.
- **`lanes import` confirmation gate.** The function took `_yes: bool` (unused); the "Lanes can execute arbitrary commands" warning was print-only. Now matches `plugin install` — non-Official sources prompt for confirmation interactively, bail with a `--yes`-required error in non-interactive mode.
- **HTTP timeouts.** Every `ureq` GitHub call in `src/github.rs` and `src/publish.rs` now uses a shared 30s `timeout_global`. A wedged endpoint no longer hangs `lanes/templates/plugins search`, `lanes import`, or the publish flow indefinitely.
- **`changelog --json` on no-tags repo.** Emits the empty envelope `{schema_version:1, action:"changelog", releases:[]}` instead of falling through to human text.
- **Docs.** `README.md` + `docs/src/templates.md` add the `kotlin-kmp` and `kotlin-ktor-api` built-ins. `CONTRIBUTING.md` drops obsolete `fledge work pr` and `fledge publish` references, notes that `fledge deps` ships in the default plugin set.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test` — 693 unit + 177 integration green, 3 consecutive runs (the `full_release_flow` flake did not reproduce)
- [x] `fledge lanes run pre-commit`
- [x] `fledge changelog --json` on a fresh tag-less git repo prints the empty envelope
- [ ] CI green

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,13 @@ All checks in the `pre-commit` lane must pass before submitting a PR.
 ### 5. Submit a Pull Request
 
 ```bash
-fledge work pr --title "Add my feature"
+# Push your branch first
+fledge work push
+
+# Open a PR (requires fledge-plugin-github)
+fledge github prs create --title "Add my feature"
+# or infer title and body from your commits:
+fledge github prs create --fill
 ```
 
 In your PR description:
@@ -85,7 +91,7 @@ Look for issues labeled [`good first issue`](https://github.com/CorvidLabs/fledg
 
 ### Templates
 
-Create new templates and publish them with `fledge publish`! See the [Template Authoring Guide](https://corvidlabs.github.io/fledge/template-authoring.html) for the full format.
+Create new templates and publish them with `fledge templates publish`! See the [Template Authoring Guide](https://corvidlabs.github.io/fledge/template-authoring.html) for the full format.
 
 ### Plugins
 
@@ -146,11 +152,13 @@ Every module has a spec in `specs/<module>/`. The spec is the source of truth fo
 
 ### Dependencies
 
-Check dependency health with:
+Dependency commands ship in `fledge-plugin-deps` (part of the default plugin
+set). Install once, then check dependency health:
 
 ```bash
-fledge deps              # list all dependencies
-fledge deps --outdated   # check for outdated deps
+fledge plugins install --defaults   # one-time, gets github/deps/metrics
+fledge deps                         # report dependency status
+fledge deps --outdated              # show outdated entries
 ```
 
 ## Release Process
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ Not every fledge user is on GitHub or runs a polyglot project. The core stays ti
 
 ## Built-in templates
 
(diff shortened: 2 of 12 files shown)


corvid-agent:

corvid-agent (approved):
Clean, well-scoped PR — six independent fixes, none of which touch happy-path behavior. Reviewed the diff in full.

**Clippy `--all-targets` + `module_inception` fix** — Correct. The old lint task silently skipped test code. Unwrapping the inner `mod tests {}` in `release/tests.rs` is the right fix; the module path was `release::tests::tests::*` which is also confusing for devs.

**Shared `CWD_LOCK` via `test_support.rs`** — This is the right structural fix. The `catch_unwind` + `resume_unwind` pattern in `with_cwd` ensures cwd is restored even on test panics, and recovering from poisoned locks (`unwrap_or_else(|e| e.into_inner())`) prevents one panicking test from cascading failures across the suite. One thing to verify: are there other tests in `lanes/tests.rs` besides `imported_lanes_get_source_tracked` that call `set_current_dir`? The diff only shows the lock added to that one test — worth a quick grep to make sure nothing else races.

**`lanes import` confirmation gate** — Good security fix. The `_yes` parameter being unused meant community lanes could execute arbitrary commands without user consent. The new behavior (interactive prompt for non-Official sources, `--yes` required in non-interactive mode) correctly mirrors `plugin install`. Placement after source resolution but before the actual import is correct.

**HTTP timeouts** — Solid. 30s `timeout_global` on all `ureq` calls prevents indefinite hangs. Creating a fresh `Agent` per function call is fine for a CLI tool — connection pooling isn't meaningful when the process exits after a handful of requests.

**`changelog --json` on tag-less repo** — Clean early return with the empty envelope. Keeps the JSON contract consistent for consumers.

**Docs** — Accurate updates to CONTRIBUTING.md, README.md, and templates.md. No issues.

LGTM — approving. The only thing I'd double-check before merge is whether `lanes/tests.rs` has any other `set_current_dir` calls that need the shared lock.

PR #324, opened 2026-05-02
fix(templates): implement copy glob, lock prompt order, add v1 compat policies

0xLeif:
## Summary

Two pre-1.0 blockers in the templates schema, plus Compatibility Policy sections locking the v1 contracts for templates and lanes.

- **`[files] copy` was documented but silently dropped.** Built-in templates (`rust-cli`, `static-site`, …) ship `copy = ["**/*.png", …]` and `create_template` scaffolds it into every new template, but `FileRules` had no `copy` field, so serde dropped it. Now wired through `render_template` + `validate` with explicit precedence: `ignore → .tera → copy → render → default-copy`. A copy-matched file bypasses Tera even when a render glob would otherwise catch it, so `render = ["**/*"]` no longer corrupts binary assets.
- **`prompts: HashMap` had non-deterministic iteration order.** Multi-prompt templates asked questions in random order across runs. Switched to `BTreeMap` — alphabetical-by-key, deterministic, no new dep.
- **Compatibility Policy sections** added to `specs/templates/templates.spec.md` (v8) and `specs/lanes/lanes.spec.md` (v20) mirroring `specs/plugin/plugin-protocol.spec.md`'s policy: additive-only sections, no field removal or retyping, locked file-rule precedence + prompt iteration order, locked parallel/fail_fast semantics, discriminator-key rule for future `Step` variants.

Both code fixes are breaking the moment 1.0 ships (HashMap iteration becomes contractual; future templates that rely on `copy` being a no-op are theoretical but possible). Landing now.

## Test plan

- [x] `cargo test` — 876 passed, 0 failed (+4 new tests)
  - `parse_manifest_with_copy_rules`
  - `prompts_iterate_in_deterministic_order`
  - `copy_glob_overrides_render_glob`
  - `tera_extension_wins_over_copy_glob`
- [x] End-to-end: `fledge templates init testproj --template static-site` renders correctly with copy globs honored
- [x] `fledge templates validate <built-in templates>` clean

### DIFF
--- a/specs/lanes/lanes.spec.md
+++ b/specs/lanes/lanes.spec.md
@@ -1,6 +1,6 @@
 ---
 module: lanes
-version: 19
+version: 20
 status: active
 files:
   - src/lanes/mod.rs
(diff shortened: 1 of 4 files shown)


corvid-agent:

corvid-agent (approved):
Reviewed the diff — approving. Clean fixes for two real pre-1.0 bugs, well tested.

**Copy glob implementation** — The precedence logic in `render_template` is correct: `ignore` short-circuits early → `.tera` extension wins (explicit render signal) → `copy` forces verbatim → `render` globs → default copy-as-bytes. The `force_copy` guard correctly gates on `!is_tera_ext` so `.tera` always takes priority. The same logic is mirrored in `validate_single` so `fledge templates validate` stays consistent with the renderer.

**`HashMap` → `BTreeMap`** — Minimal, correct fix. Alphabetical-by-key is stable and needs no new dependency. The contract test (`prompts_iterate_in_deterministic_order`) locks this.

**Tests** — The four new tests hit the right cases:
- `parse_manifest_with_copy_rules` — basic deserialization
- `prompts_iterate_in_deterministic_order` — contract lock for iteration order
- `copy_glob_overrides_render_glob` — the critical behavioral case (broad `render = ["**/*"]` doesn't corrupt binary assets)
- `tera_extension_wins_over_copy_glob` — `.tera` override path

**Compatibility policies** — Well-structured, practical. The prompt-key naming guidance (`01_name`, `02_description`) for UX ordering is a nice touch given the alphabetical lock.

One minor note for future: the `force_copy` + `should_render` logic is duplicated between `templates.rs` and `validate.rs`. Reasonable trade-off at this stage — just worth watching if more glob categories (e.g. `binary`, `executable`) get added per the spec.

Ship it. 🚢

PR #325, opened 2026-05-02
fix(introspect): propagate inherited global args to descendant subcommands

0xLeif:
## Summary

`fledge introspect --json` is the discovery API agents use to enumerate the CLI surface. Previously a clap `global = true` arg declared on a parent command was emitted only on that parent — not on the subcommands that actually accept it. So an agent reading `plugins.list.args` saw an empty array, even though `--json` (declared on `Plugins` with `global = true`) and `--non-interactive` (declared on the root) both work there.

Now each node's `args` reflects the **complete set of flags accepted at that level**, with inherited args retaining `global: true` so agents can distinguish them from locally-declared ones if they care. A child re-declaring an inherited arg by name keeps its own copy — local wins, no duplicates.

The wire format is unchanged: `global: true` was always part of the v1 introspect schema, this fix uses it as intended. `INTROSPECT_SCHEMA_VERSION` stays at 1.

### Before (root-only globals)
```
$ fledge introspect --json | jq '.subcommands[] | select(.name=="plugins").subcommands[] | select(.name=="list").args'
[]
```

### After
```
[
  { "name": "json", "long": "json", "global": true, ... },
  { "name": "non_interactive", "long": "non-interactive", "global": true, ... }
]
```

## Changes
- `src/introspect.rs`: `build_tree` threads inherited globals through recursion, dedupes by name
- `specs/introspect/introspect.spec.md` → v3: invariant 5 flipped, change-log entry
- `AGENTS.md`: row clarified — args list now includes inherited globals
- `src/snapshots/...introspect_schema.snap` regenerated

## Test plan
- [x] `cargo test` — **879 passed, 0 failed** (+3 new tests)
  - `global_args_propagate_to_subcommands`
  - `globals_propagate_through_multiple_levels`
  - `child_redeclaration_does_not_duplicate_inherited_arg`
- [x] Live binary: `fledge plugins list` introspect node now reports both `--json` and `--non-interactive` with `global: true`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,7 +49,7 @@ Specs (`specs/<name>/*.spec.md` and companion files) are the source of truth for
 
 | Command | What you get | Use when |
 |---------|-------------|----------|
-| `fledge introspect --json` | `{schema_version: 1, name, about, aliases, args: [...], subcommands: [...]}`. Each subcommand recursively has the same shape; each arg has `name, long?, short?, aliases, help, required, takes_value, value_name, global?` | First contact with fledge |
+| `fledge introspect --json` | `{schema_version: 1, name, about, aliases, args: [...], subcommands: [...]}`. Each subcommand recursively has the same shape; each arg has `name, long?, short?, aliases, help, required, takes_value, value_name, global?`. Each node's `args` is the **complete set of flags accepted at that level**, including inherited globals from ancestors (marked `global: true`) — no need to walk up the parent chain | First contact with fledge |
 | `fledge spec list --json` | `{schema_version: 1, action: "spec_list", specs: [{name, version, status, sections, companions, ...}]}` | Orienting to a new codebase |
 | `fledge spec show <name> --json` | `{schema_version: 1, action: "spec_show", spec: {name, version, status, sections, companions, ...}}` | Need structured view of one module |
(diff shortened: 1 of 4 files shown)


corvid-agent:

corvid-agent (approved):
Reviewed the diff — clean fix, approving.

**Implementation** (`src/introspect.rs`): The `build_tree_with_inherited` refactor is straightforward. Threading `inherited: &[ArgNode]` through recursion is the right approach. Dedup-by-name with local-wins semantics is correct, and the `next_inherited` collection correctly picks up both newly-declared globals and already-inherited ones for further propagation. `Clone` on `ArgNode` is the right call — it's a small struct.

**Tests**: The three new cases cover the important contract points well — basic propagation, multi-level depth, and child redeclaration shadow behavior. Using purpose-built test CLIs (`DeepCli`, `ShadowCli`) keeps them readable and isolated from the main `TestCli`.

**Spec**: Invariant 5 flip is clearly documented, changelog entry is thorough. Keeping `INTROSPECT_SCHEMA_VERSION` at 1 is the right call since `global: true` was always part of the v1 shape — this is a behavioral fix, not a schema change.

**Snapshot**: The growth is mechanical and expected — every subcommand now carries the inherited globals.

No issues found. Nice fix, Leif.

— CorvidAgent (Rook)

PR #326, opened 2026-05-02
fix: third-review pre-1.0 polish (security docs, scope validation, stderr redaction)

0xLeif:
## Summary

Five small fixes from the independent third reviewer's findings (D1, D2, D3, D5, D6). The four review blockers (B, C1, C2, C3) were already in main via #324 and #325 — the reviewer reviewed against `9d6636b`, two PRs behind.

| ID | Severity | Fix |
|----|----------|-----|
| **D1** | Doc | SECURITY.md plugins section now explicitly notes that granting `exec` capability = full shell access within the sandbox |
| **D5** | Doc | SECURITY.md supported-versions table no longer claims 1.0 support before 1.0 exists; adds 0.17.x "best effort" row |
| **D6** | Doc | AGENTS.md notes `work status --json` migrated from schema_version 1 → 2 in 0.16 |
| **D2** | Defensive | `redact_secrets()` strips `Authorization:`, `x-access-token:`, `Bearer`, and credentialed URLs from raw subprocess stderr before it lands in user-facing error messages. Wired into `remote.rs` (the only authenticated-git site). The token is sent via env headers (not URL), so this closes a future-proofing gap rather than a live leak |
| **D3** | Security (low) | `work commit --ai --scope <s>` validates `<s>` against `[A-Za-z0-9_-]{1,64}` before interpolating into the LLM prompt or commit message. Rejects whitespace, shell metacharacters, template syntax, or anything that reads as instructions to the model |

### Deliberately not included

**D4 (separate `--trust-hooks` consent for remote-template hooks, distinct from `--yes`)** is a UX trade-off worth discussing before 1.0 lock-in. Currently `--yes` (and `FLEDGE_NON_INTERACTIVE=1`) auto-grant remote-template `post_create` hook execution. SECURITY.md documents this accurately. The reviewer suggests requiring an explicit `--trust-hooks` flag for remote templates — that's a behavior change to `--yes` semantics that could break existing CI flows. Flagged for separate decision.

## Test plan
- [x] `cargo test` — **888 passed, 0 failed** (+9 new tests)
  - 6 redaction tests (Authorization header, x-access-token, URL creds, Bearer, clean passthrough, case-insensitive)
  - 3 commit-scope tests (accepts normal, rejects injection vectors, length cap)
- [x] `fledge spec check` — 29 specs, 0 errors, 0 warnings (work spec → v12)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -131,7 +131,7 @@ When non-interactive mode is active, every command that would otherwise prompt b
 
 | Command | Effect |
 |---------|--------|
-| `fledge templates init` | Skip template-variable prompts (uses detected defaults) |
+| `fledge templates init` | Skip template-variable prompts (uses detected defaults). For **local** templates this also auto-confirms `post_create` hooks. For **remote** templates it does **not** — pass `--trust-hooks` (or set `FLEDGE_TRUST_HOOKS=1`) to authorize hooks from a third-party source. Without it, hooks are skipped in non-interactive mode and the rest of init still succeeds (`hooks_run: false` in the JSON envelope) |
 | `fledge templates create` | Skip name/description/type prompts |
 | `fledge ai use` | Errors with a clear "pass provider+model" message. No hang |
 | `fledge work commit` | Skip the interactive message prompt (requires `-m` or `--ai`) |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,7 +92,7 @@ Specs (`specs/<name>/*.spec.md` and companion files) are the source of truth for
 | `fledge work start <name> --json` | `{schema_version: 1, action: "work_start", branch, base, type, prefix, issue}`. `issue` is `null` when no `--issue` flag was passed; all other fields always present | Branch scripting |
 | `fledge work commit --json` | `{schema_version: 1, action: "work_commit", hash, message, branch}`. Commit hash to report back | After writing code |
 | `fledge work push --json` | `{schema_version: 1, action: "work_push", branch, remote, force}`. Confirms the push | After committing |
-| `fledge work status --json` | `{schema_version: 2, action: "work_status", branch, default, ahead, behind, dirty}`. `dirty` is uncommitted file count; no PR field | Pre-action sanity check |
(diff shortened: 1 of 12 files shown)


corvid-agent:

corvid-agent (dismissed):
**Rook — security/architecture review**

Clean, well-scoped PR. All five fixes are defensive and correctly targeted. Approving.

### Scope validation (D3) ✅

`validate_commit_scope` is placed exactly right — at the top of `commit()` in `work.rs`, before any git operations or LLM prompt interpolation. The character-by-character allowlist (`is_ascii_alphanumeric || '-' || '_'`) with a 64-char cap is strict and correct. The test vectors cover LLM prompt injection ("ignore previous instructions"), shell metacharacters, template syntax, and path traversal — good adversarial thinking.

### Secret redaction (D2) ✅

`redact_secrets()` covers the four credential patterns that matter for git subprocess stderr: `Authorization:` headers, `x-access-token:` (GitHub's PAT header), URL-embedded credentials, and `Bearer` tokens. The case-insensitive matching and end-of-line capture for header values (not just `\S+`) is the right call — avoids partial redaction of multi-token auth values like `Basic <base64>`.

Wiring it only into `remote.rs::clone_repo` matches the stated threat model (only authenticated-git site). Correct scope.

**One minor nit:** the four `Regex::new().unwrap()` calls compile on every invocation. Since this is error-path-only code, it's fine for now, but if `redact_secrets` ever gets called in a hot path, consider `std::sync::LazyLock` or `once_cell::Lazy` to compile once.

**Cosmetic note:** the Bearer regex `[^\s\n]+` will consume trailing punctuation (e.g., `(Bearer token123)` → `(Bearer [REDACTED]` — eats the closing paren). Not a correctness issue since the output is already a redacted error message, just slightly odd formatting. Low priority.

### Documentation (D1, D5, D6) ✅

- SECURITY.md `exec` capability warning is clear and accurate — calling out "no shell-metacharacter filtering" is the right thing to document
- Supported-versions table no longer overclaims 1.0; 0.17.x best-effort row is sensible pre-1.0
- AGENTS.md v1→v2 migration note with "pin to ≥ 0.16" guidance is helpful for downstream consumers

### D4 deferral

Agree this deserves separate discussion. The `--yes` vs `--trust-hooks` split has CI-flow implications that shouldn't be buried in a polish PR.

LGTM. Ship it.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #329, opened 2026-05-02
fix(introspect): correct takes_value detection — third-review last-mile blocker

0xLeif:
## Summary

Independent third-pass review caught a 1.0 blocker in the introspect output that both prior reviews missed.

The previous probe in `src/introspect.rs:109-112`:

```rust
let takes_value = arg
    .get_num_args()
    .map(|n| n.takes_values())
    .unwrap_or(false);
```

returns `None` for the vast majority of clap-derive args — `get_num_args()` is only `Some` when the spec explicitly writes `#[arg(num_args = ...)]`. The `unwrap_or(false)` then made every `Option<String>`, `PathBuf`, `String` positional, and `Vec<String>` flag look like a boolean.

### Verified pre-fix on the live binary

```text
templates init: --template/--output/--author/--org → takes_value: false
ask:            --with-specs/--provider/--model    → takes_value: false
work commit:    --message/--scope/--commit-type    → takes_value: false
```

`value_name` was also suppressed (the line 133 guard depends on `takes_value`), so agents got zero type information on the args that most need it. An agent reading `templates init.args` could not distinguish `--no-git` (flag) from `--template <TEMPLATE>` (value option). The whole point of `fledge introspect --json` is to let agents generate correct invocations from the JSON tree; this bug made that impossible for nearly every value-taking flag.

### Why this had to ship pre-1.0

The v1 introspect schema locks `takes_value`'s semantics. Releasing 1.0 with these wrong values, then fixing later, is a soft contract break — any agent that adapted to the broken values has to update.

### Fix

Positive-match on `ArgAction::Set | Append` instead of the broken `get_num_args()` probe. Verified post-fix on the same commands:

```text
templates init: --template/--output/--author/--org → takes_value: true (with value_name)
ask:            --with-specs/--provider/--model    → takes_value: true
work commit:    --message/--scope/--commit-type    → takes_value: true
```

Boolean flags (`--no-git`, `--yes`, `--json`, `--trust-hooks`, etc.) correctly stay `takes_value: false`. The wire format and `INTROSPECT_SCHEMA_VERSION` are unchanged — same fields, same shape, just correct values. Snapshot regenerated.

Positive-matching (rather than negative-matching against the boolean actions) is more conservative: a future clap `ArgAction` variant defaults to "doesn't take a value" rather than silently flipping the contract.

Introspect spec bumped v3 → v4 with a change-log entry.

## Test plan

- [x] `cargo doc --no-deps` clean
- [x] `cargo test` — **890 passed, 0 failed, 1 ignored**
- [x] Manual verification: spot-checked `templates init`, `ask`, `work commit` args

### DIFF
--- a/specs/introspect/introspect.spec.md
+++ b/specs/introspect/introspect.spec.md
@@ -1,6 +1,6 @@
 ---
 module: introspect
-version: 3
+version: 4
 status: active
 files:
   - src/introspect.rs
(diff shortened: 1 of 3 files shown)


corvid-agent:

corvid-agent (approved):
Good catch — this is a legitimate 1.0 blocker. The fix is correct and well-reasoned.

**Why the old code was wrong:** `get_num_args()` returns `None` for the vast majority of clap-derive args (only `Some` when the user explicitly writes `#[arg(num_args = ...)]`). The `unwrap_or(false)` then collapsed every `Option<String>`, `PathBuf`, `String`, and `Vec<String>` arg to `takes_value: false`, making them indistinguishable from boolean flags in the introspect output.

**Why the fix is correct:** `ArgAction::Set` is the default action for value-taking args, `ArgAction::Append` covers `Vec<T>` args. Boolean flags use `SetTrue`/`SetFalse`/`Count`, and utility actions use `Help`/`Version` — none of which match. Positive-matching is also the right conservatism: a hypothetical future `ArgAction` variant defaults to "doesn't take a value" rather than silently flipping the contract.

**Snapshot review:** All changes are `takes_value: false` → `true` with accompanying `value_name` fields, exclusively on args that clearly accept values (provider, model, name, path, query, etc.). Boolean flags (`--json`, `--yes`, `--no-git`, `--install`, `--trust-hooks`) are absent from the diff, confirming they're unaffected.

**One note on the comment:** The 8-line explanation in `build_arg` is longer than typical, but justified here — this exact spot has been wrong twice, and the comment explains *why* the obvious-looking `get_num_args()` approach doesn't work. Prevents the third time.

Spec bump v3→v4 is appropriate. Ship it.

PR #330, opened 2026-05-02
fix(release): dry-run files_to_bump must include [release].files extras

0xLeif:
## Summary

Caught while running `fledge release major --dry-run --json` to prep the actual 1.0.0 release. The envelope reported:

```json
"files_to_bump": ["Cargo.toml"]
```

even though `[release].files = ["flake.nix"]` is set in `fledge.toml` and a real release would also write `flake.nix`. The discrepancy: `detect_version_files()` (dry-run preview path) only looked at hardcoded language candidates, while `bump_version_files()` (real write path) also processed `[release].files` from `fledge.toml`.

This is a v1 contract bug. The dry-run envelope is part of the release JSON contract (`schema_version: 1`). Agents reading it to predict a release would get wrong info, and shipping 1.0 with this baked in means fixing it later changes observable behavior — exactly the soft contract break we've been hunting all week.

## Fix

Mirror the bumper's `[release].files` logic in `detect_version_files`:
- Read `[release].files` from `fledge.toml`
- Skip entries already in the language-candidate list (no duplicates)
- Skip non-existent files (can't be bumped)
- Skip files with no parseable `version = "X.Y.Z"` line (the real bumper wouldn't touch them either)

### Verified on live binary

Before:
```json
"files_to_bump": ["Cargo.toml"]
```

After:
```json
"files_to_bump": ["Cargo.toml", "flake.nix"]
```

## Test plan

- [x] `cargo doc --no-deps` clean
- [x] `cargo test` — **893 passed, 0 failed, 1 ignored** (was 890, +3 new)
  - `detect_version_files_includes_release_files_extras`
  - `detect_version_files_skips_release_extras_without_version_line`
  - `detect_version_files_skips_missing_release_extras`
- [x] `fledge spec check` — 29 specs, 0 errors, 0 warnings (release spec → v5)
- [x] Live binary dry-run shows correct file list

### DIFF
--- a/specs/release/release.spec.md
+++ b/specs/release/release.spec.md
@@ -1,6 +1,6 @@
 ---
 module: release
-version: 4
+version: 5
 status: active
 files:
   - src/release/mod.rs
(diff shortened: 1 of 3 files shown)


corvid-agent:

corvid-agent (approved):
**LGTM — clean fix for a real contract bug.**

The approach is correct: mirror `bump_version_files`'s `[release].files` handling in `detect_version_files` so dry-run and real-run report the same file set. The three guard clauses (already in list, file missing, no version line) match the real bumper's behavior, which is exactly what keeps the contract honest.

**Code notes:**

- The `Regex::new(...).unwrap()` is safe since it's a constant pattern, and it's constructed once before the loop — no per-iteration compile cost.
- Error handling via `if let Ok(...)` for the `fledge.toml` read/parse is the right call — no `fledge.toml` means no extras, not a crash.
- The duplicate-skip check (`files.iter().any(|f| f == name)`) is case-sensitive, which is correct for file paths.

**Minor observation (non-blocking):** The version-line regex (`(?m)(version\s*[=:]\s*["']?)(\d+\.\d+\.\d+)`) is duplicated between `detect_version_files` and `bump_version_files`. Extracting it as a module-level constant would prevent future drift. Not worth blocking on — just worth noting for a post-1.0 cleanup pass.

**Tests:** The three new tests cover the contract-critical cases well: extras included, no-version-line skipped, missing-file skipped. Good coverage.

Spec bump to v5 with accurate changelog. Ship it. 🚢

PR #340, opened 2026-05-02
docs: add consolidated fledge.toml reference page

0xLeif:
## Summary
The full schema for `fledge.toml` was scattered: a brief mention in `configuration.md`, tasks/lanes in `lanes.md`, and the `[release].files` key wasn't documented anywhere. This PR adds a single canonical reference and corrects some drift in `plugins.md`.

- **New** `docs/src/fledge-toml.md` — every section (`[tasks]`, `[lanes]`, `[release]`, `schema_version`, imported `.fledge/lanes/*.toml`), every key with type/default/notes, behavior reference, full worked example, and validation commands.
- `SUMMARY.md` — wire the new page into the TOC under "Run: Tasks and Lanes".
- `configuration.md` — replace the thin "see lanes.md and plugins.md" stub with a link to the new ref.
- `lanes.md` — cross-link in the Related section.
- `plugins.md` — two corrections grounded in `src/plugin/mod.rs` and `src/protocol/mod.rs`:
  - Document the previously-missing `protocol` field on `[plugin]` (the `"fledge-v1"` opt-in).
  - Replace the inaccurate "Message Types" table. The old table claimed `exec_result` / `store_ack` / `load_result` / `metadata_result` envelopes that don't exist in the source — the protocol uses a generic `response` with an echoed `id` for everything. It also said `output` "terminates plugin", but `output` is fire-and-forget and just prints the text. The new table matches `OutboundMessage` and the dispatch in `run_message_loop`.

## Why
- Authors writing `fledge.toml` had to read three pages and a source struct to learn the schema.
- The `[release].files` key was undocumented but is real — `fledge release` reads it, and `release --dry-run` reports against it.
- Plugin authors looking at the message-types table got incorrect reply shapes, which would not match the actual wire protocol.

## Notes
- Audit-grounded: I read `src/run.rs` (`TaskConfig`), `src/lanes/mod.rs` (`LaneDef`, `Step`, `ParallelItem`), `src/release/bump.rs` (`[release]` parsing and the canonicalization invariant), `src/plugin/mod.rs` (`PluginManifest`), and `src/protocol/mod.rs` (`OutboundMessage`).
- Links to the existing canonical `specs/plugin/plugin-protocol.spec.md` for protocol details rather than introducing a new `plugin-protocol.md` page.
- Imported-lanes section documents `.fledge/lanes/*.toml` auto-load behavior I found in `src/lanes/mod.rs::load_lane_config`.

## Test plan
- [ ] `mdbook build docs` succeeds (couldn't run locally — `mdbook` not installed in my sandbox)
- [ ] mdbook navigation shows the new page under "Run: Tasks and Lanes"
- [ ] Confirm cross-links from `configuration.md`, `lanes.md`, and the new page all resolve
- [ ] Spot-check the worked example parses with `cargo run -- lanes validate` against a fixture project

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -9,6 +9,7 @@
 - [Scaffold: Templates](./templates.md)
   - [Template Authoring Guide](./template-authoring.md)
 - [Run: Tasks and Lanes](./lanes.md)
+  - [`fledge.toml` Reference](./fledge-toml.md)
 - [Spec: Spec-sync](./spec.md)
 - [AI: Ask and Review](./review.md)
 - [Ship: Branch, PR, Release](./ship.md)
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -152,11 +152,13 @@ If neither env var nor config is set, fledge falls back to `gh auth token` (GitH
 
 ## Project Configuration (fledge.toml)
 
-Per-project settings live in `fledge.toml` in your project root. This file defines tasks, lanes, and project metadata. It's created by `fledge run --init` or `fledge templates init`.
+Per-project settings live in `fledge.toml` in your project root. This file defines tasks, lanes, and release behavior. It's created by `fledge run --init`, `fledge lanes init`, or `fledge templates init`.
 
-For task and lane configuration, see:
-- [Run: Tasks and Lanes](./lanes.md) — defining lanes, step types, parallel groups, importing community lanes
-- [Extend: Plugins](./plugins.md) — extending fledge with community plugins
+For the full schema (every section, every key, every default), see the [`fledge.toml` Reference](./fledge-toml.md).
+
+Topical guides:
+- [Run: Tasks and Lanes](./lanes.md) — defining tasks/lanes, step types, parallel groups, importing community lanes
+- [Extend: Plugins](./plugins.md) — `plugin.toml` and the plugin ecosystem
 
 ## Priority Order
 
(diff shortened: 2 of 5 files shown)


corvid-agent:

corvid-agent (dismissed):
Clean, thorough docs PR. Reviewed the full diff — approving.

**What's good:**

- The new `fledge-toml.md` is well-structured: every section/key/default in one place with a full worked example. This is the page I'd want to land on as a new fledge user.
- The plugin protocol table correction is important. The old `exec_result` / `store_ack` / `load_result` / `metadata_result` envelopes would have plugin authors writing code against shapes that don't exist on the wire. The generic `response` with echoed `id` matches what you'd actually see.
- Good call adding the `protocol` field to the `[plugin]` manifest table — that was a real gap.
- Cross-links are wired up in all directions (configuration → ref, lanes → ref, ref → everything).
- Behavior reference section captures the subtleties (project precedence disabling auto-detect, empty sections being errors, path escape rejection) that would otherwise only be discoverable by reading source.

**One nit (non-blocking):**

- In the `[release]` section, the regex `version\s*[=:]\s*["']?(\d+\.\d+\.\d+)` is great to document. Might be worth noting whether it matches only the first occurrence per file or all of them — but that can be a follow-up.

LGTM. The test plan items (mdbook build, nav check, cross-link resolution, validation against fixture) all look right for this scope.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #352, opened 2026-05-03
fix(lint): silence clippy::const_is_empty on DEFAULT_PLUGINS

0xLeif:
## Summary

Rust 1.91 added the `clippy::const_is_empty` lint which flagged `assert!(!DEFAULT_PLUGINS.is_empty())` in `src/plugin/tests.rs` because the slice's emptiness is known at compile time. With `-D warnings` in CI, this blocks the lint step.

Replaced the runtime assertion with a compile-time `const _: () = assert!(...)` outside the test fn. Same invariant, no clippy noise, fails the build at compile time instead of at test time if someone empties the const.

Blocks 1.1.1 release, lint is currently red on main.

## Test plan

- [x] `fledge lanes run pre-commit` (fmt + lint + test + spec-check, all green)
- [x] `cargo clippy --tests -- -D warnings` clean

### DIFF
--- a/src/plugin/tests.rs
+++ b/src/plugin/tests.rs
@@ -119,9 +119,13 @@ fn resolve_plugin_source_dir_handles_non_symlink_path() {
     assert_eq!(resolved, expected);
 }
 
+const _: () = assert!(
+    !DEFAULT_PLUGINS.is_empty(),
+    "DEFAULT_PLUGINS must not be empty"
+);
+
 #[test]
 fn default_plugins_are_well_formed() {
-    assert!(!DEFAULT_PLUGINS.is_empty());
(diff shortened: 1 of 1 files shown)


corvid-agent:

corvid-agent (approved):
Clean fix. Moving the emptiness check from a runtime `assert!` inside the test to a `const _: () = assert!(...)` at module scope is strictly better — same invariant, enforced at compile time instead of test time, and silences the `clippy::const_is_empty` lint without an allow attribute.

The remaining `default_plugins_are_well_formed` test still validates structure via `parse_source_ref`, so no coverage is lost.

LGTM — unblocks the 1.1.1 release.

PR #363, opened 2026-05-04
feat(lanes): retry_delay, Windows job kill, thread-safe when tests

0xLeif:
## Summary

Three follow-up items from #362's review thread that I noted as non-blocking but worth landing.

- **`retry_delay` step option** (configurable inter-attempt sleep, default 1s, supports `retry_delay = 0` for immediate retry)
- **Windows process tree kill on timeout** via Job Object + `TerminateJobObject` — mirrors the v23 Unix `process_group` + `killpg` fix, so `cmd /c "a & b"` no longer leaks grandchildren
- **Thread-safe `when` tests** — `evaluate_when_with` now takes a closure so tests pass a `HashMap` instead of mutating `std::env::set_var` (edition-2024 prep)

## Test plan

- [x] `cargo test --bin fledge` — 834 pass (4 new)
- [x] `cargo test --test lanes` — 9 pass
- [x] `cargo clippy -- -D warnings` clean (Linux + `x86_64-pc-windows-gnu` cross)
- [x] `cargo run -- spec check` — 30/30 green
- [x] Manually verified `retry_delay = 0` (16ms total, no inter-attempt sleep) and `retry_delay = 3` (6.03s total = 2 × 3s)
- [x] Cross-compiled to `x86_64-pc-windows-gnu` to validate Windows path

## Behavioral examples

```toml
# Immediate retry, no delay
{ run = "curl health", retries = 3, retry_delay = 0 }

# Backoff-style longer wait
{ task = "deploy", retries = 2, retry_delay = 30 }

# Default — 1s delay, unchanged
{ run = "flaky-test", retries = 3 }
```

Spec bumped 23 → 24.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,5 +51,4 @@ libc = "0.2.185"
 windows-sys = { version = "0.59", features = [
     "Win32_Foundation",
     "Win32_System_JobObjects",
-    "Win32_System_Threading",
 ] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,3 +46,10 @@ tempfile = "3"
 
 [target."cfg(unix)".dependencies]
 libc = "0.2.185"
+
+[target."cfg(windows)".dependencies]
+windows-sys = { version = "0.59", features = [
+    "Win32_Foundation",
+    "Win32_System_JobObjects",
+    "Win32_System_Threading",
+] }
--- a/specs/lanes/lanes.spec.md
+++ b/specs/lanes/lanes.spec.md
@@ -1,6 +1,6 @@
 ---
 module: lanes
-version: 23
+version: 24
 status: active
 files:
   - src/lanes/mod.rs
@@ -141,7 +141,8 @@ steps = ["deps-audit", "license-check", "security-scan"]
 |--------|------|-------------|
 | `when` | `string` | Condition: skip step if not met. Supports `VAR` (set & non-empty), `VAR=value` (equals), `!VAR` (not set/empty), `!VAR=value` (not equals). Multiple comma-separated conditions are AND'd |
 | `timeout` | `integer` | Per-attempt timeout in seconds. Each retry attempt gets a fresh deadline. The deadline covers the entire step including task deps. Process is killed on timeout |
-| `retries` | `integer` | Number of retry attempts after failure (0 = no retries, default). Total attempts = retries + 1. A 1-second delay is inserted between attempts |
+| `retries` | `integer` | Number of retry attempts after failure (0 = no retries, default). Total attempts = retries + 1. The delay between attempts is configurable via `retry_delay` (defaults to 1 second) |
+| `retry_delay` | `integer` | Seconds to sleep between retry attempts. Defaults to 1. Set to 0 for immediate retry. Only meaningful when `retries > 0` |
 
 ```toml
 # Examples of step options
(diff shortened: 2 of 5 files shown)


corvid-agent:

corvid-agent (dismissed):
Clean PR — three well-scoped follow-ups from the v23 review thread.

## `retry_delay`

Straightforward addition. `u64` type prevents negative values, `sleep(0)` for immediate retry is a harmless yield, and the `DEFAULT_RETRY_DELAY_SECS` constant keeps the default centralized. Backwards-compatible since unspecified retries still sleep 1s.

## Windows Job Object

Good approach. Key observations:

- **Race window** between `spawn` and `AssignProcessToJobObject` is acknowledged and acceptable — the alternative (`CREATE_SUSPENDED` + `ResumeThread`) adds substantial complexity for negligible benefit in typical shell invocations.
- **Handle lifecycle** is correct: closed on all paths (normal exit, timeout, and every setup-failure early-return).
- **`KILL_ON_JOB_CLOSE`** is a nice property — also reaps children if fledge itself crashes unexpectedly.
- **Graceful degradation** to `child.kill()` when job setup fails (resource exhaustion) is the right fallback.

One minor note: the old `#[cfg(not(unix))]` catch-all is now `#[cfg(windows)]` specifically. If anyone ever targets a platform that is neither Unix nor Windows, there's no timeout-kill path. Unlikely to matter in practice, but worth documenting as a known limitation if you want to be thorough.

## Thread-safe `when` tests

The `evaluate_when_with` closure injection is the correct pattern for edition-2024 prep. The `matches!` rewrites are more idiomatic than the old `match` blocks, and the new `evaluate_when_var_set_but_empty` test adds coverage for a previously-untested edge case. The single `evaluate_when_real_env_smoke` test keeps the actual codepath honest without needing `unsafe set_var`.

LGTM — approve.

0xLeif:
## Addressed in `5699c2f`

Thanks for the careful read — three of the four landed verbatim, one was a real catch I missed:

**1. KILL_ON_JOB_CLOSE (the broader-than-timeout reaping behavior)** — fair call. With the flag set, a successful step that intentionally backgrounds a daemon would have its daemon killed when the job handle drops, which doesn't match the Unix path (Unix only kills on `killpg`-on-timeout). Dropped the flag and the `SetInformationJobObject` call entirely. The only kill site is now `TerminateJobObject` on the timeout branch.

**2. Handle leak on `?` propagation** — fixed via a `JobHandle` RAII guard. `Drop` calls `CloseHandle` on every exit path including the error return from `child.try_wait()`. All the manual close calls are gone.

**3. Unused `Win32_System_Threading` feature** — you were right and I missed this. Once I dropped the limit-flag setup in (1), `JOBOBJECT_EXTENDED_LIMIT_INFORMATION` and `SetInformationJobObject` are gone, and nothing behind that feature is referenced anymore. Verified by re-running `cargo check --target x86_64-pc-windows-gnu` after pulling the feature — still compiles clean.

**4. 800ms test bound flake risk** — bumped to 1500ms. Default behavior with `retries = 2` adds 2s of sleep, so 1500ms still cleanly distinguishes "delay was honored" from "delay was skipped" without tripping on a slow CI runner.

Verified locally on both Linux and `x86_64-pc-windows-gnu` cross — clippy + fmt + tests clean.

PR #368, opened 2026-05-06
fix(protocol): prevent stdin inheritance in exec child processes

corvid-agent:
## Summary
- Exec'd child processes inherited the parent's stdin, causing nested plugin invocations (e.g., `fledge memory save` calling `fledge sql query` internally) to race on the same pipe and drop messages after ~4 sequential execs.
- Fix: set `stdin(Stdio::null())` on exec child spawns. Exec commands receive input via their command string, not stdin — they should never inherit the plugin protocol pipe.
- Not a breaking change — no existing plugin reads from stdin in an exec'd process.

## Test plan
- [x] `cargo build` — compiles cleanly
- [x] `cargo test` — all 23 tests pass
- [x] `cargo clippy -- -D warnings` — no warnings
- [ ] Manual: `fledge memory save --key test --value hello` (triggers nested `fledge sql migrate` + `fledge sql query`)


corvid-agent:

corvid-agent:
CI all green — thanks Corvin! 🪶 Ready to merge whenever Leif gives the nod.

corvid-agent:

corvid-agent:

corvid-agent:

PR #376, opened 2026-05-07
Update README.md

0xLeif:


corvid-agent:

corvid-agent (approved):
LGTM. Adding `fledge lanes init` before `fledge lanes run ci` in the quickstart is the right call — without it, new users following the example would hit an error on the `lanes run` step. Clean one-liner fix.

Issue #383, opened 2026-05-12
`fledge spec new` rejects names with path separators


0xLeif:
## Summary

`fledge spec new <name>` errors out when `<name>` contains a `/`, blocking the natural domain-grouped spec layout users describe in their project docs.

## Repro

```bash
$ fledge spec new game/board --ni
error: Invalid module name 'game/board': may not contain path separators or '..'
```

## Source

`src/spec/mod.rs`:

```rust
if name.contains('/') || name.contains('\\') || name.contains("..") {
    bail!("Invalid module name '{name}': may not contain path separators or '..'");
}
```

The `..` rejection is fine (path traversal). The `/` and `\\` rejection is what blocks legitimate domain-grouped layouts.

## Expected

Either:

1. Accept nested names so they produce a nested layout, e.g. `specs/game/board/board.spec.md`, with the registry key `game/board` (or `game.board`).
2. Or document the flat-only model up front and add a `--group` flag (`fledge spec new board --group game`) that achieves the same outcome.

## Why this matters

When bootstrapping a project, the natural way to describe specs in `AGENTS.md` / `README.md` is by domain:

```
specs/game/board.spec.md
specs/game/pieces.spec.md
specs/network/websocket.spec.md
specs/auth/auth.spec.md
specs/protocol/messages.spec.md
```

Users following that doc hit the path-separator error immediately on the first `fledge spec new`. The current workaround is to use flat names (`board`, `websocket`, `messages`), which loses the domain grouping and forces a doc rewrite to match `specs/<name>/<name>.spec.md`.

## Workaround

Use flat names and accept losing the domain prefix:

```bash
fledge spec new board       # → specs/board/board.spec.md
fledge spec new websocket   # → specs/websocket/websocket.spec.md
```

## Environment

- `fledge` (latest, via cargo)
- macOS (Darwin 25.2.0)

0xLeif:
Fixed in #384.

Issue #382, opened 2026-05-10
Skip workspace-managed plugins during `fledge plugins update`


corvid-agent:
## Problem

Plugins that live inside a monorepo (e.g. Merlin's internal plugins under `plugins/`) are registered with `source = "CorvidLabs/merlin"` but aren't standalone repos. Running `fledge plugins update` tries to `git pull` in their install directory, which fails:

```
⚠️ fledge-plugin-search — git pull failed, try reinstalling with fledge plugin install CorvidLabs/merlin --force
⚠️ fledge-plugin-specsync — git pull failed
⚠️ fledge-plugin-files — git pull failed
⚠️ fledge-plugin-git — git pull failed
⚠️ fledge-plugin-shell — git pull failed
⚠️ fledge-plugin-web — git pull failed
⚠️ fledge-plugin-snapshots — git pull failed
```

These plugins are managed by the host project's init/build step (e.g. `merlin init`), not by fledge's git-based update mechanism.

## Proposed Solution

Add a `local = true` flag (or `source = "local"`) to `plugins.toml` entries that tells `fledge plugins update` to skip them silently. The registration code in the host project would set this flag when installing workspace plugins.

Alternatively, detect that the plugin directory has no `.git` and skip gracefully with an info message instead of a warning.

## Workaround

Run the host project's init command (`merlin init`) to rebuild and re-symlink workspace plugins. Ignore the warnings from `fledge plugins update`.

0xLeif:
Fixed in #384.

Issue #379, opened 2026-05-08
No `ai.claude.api_key` config key — Claude provider requires undocumented env var


corvid-agent:
## Feature / Bug

The Ollama provider has `ai.ollama.api_key` in the config schema, but the Claude provider has no equivalent `ai.claude.api_key`. Users who want to use the Claude provider must set `ANTHROPIC_API_KEY` as an environment variable, but this isn't surfaced anywhere in fledge's config system.

### Steps to reproduce

```bash
fledge config set ai.claude.api_key "sk-ant-..."
# error: Unknown config key 'ai.claude.api_key'. Valid keys: ...
```

Valid Claude config keys are limited to just `ai.claude.model`.

### Expected behavior

Either:
1. Add `ai.claude.api_key` to the config schema (consistent with `ai.ollama.api_key`), OR
2. `fledge ai status` should show where the API key is expected when using Claude provider: `"API Key: from ANTHROPIC_API_KEY env var"` or `"API Key: (not set — export ANTHROPIC_API_KEY to use Claude)"`
3. `fledge config list` should document the required env var for the Claude provider

### Impact

Users switching from Ollama to Claude don't know how to configure the API key since the config system doesn't support it and doesn't point them to the env var.

### Environment

- fledge 1.3.1
- macOS (Darwin 25.3.0)

---
🤖 Agent: CorvidAgent | Model: Opus 4.6
👤 Requested by: @Kyntrin

0xLeif:
Fixed in #384.

Issue #378, opened 2026-05-08
`OLLAMA_HOST` env var overrides config without clear feedback in error messages


corvid-agent:
## Bug

When `OLLAMA_HOST` is set as an environment variable, it overrides the `ai.ollama.host` config value. This means even after changing the host via `fledge config set`, the env var silently takes precedence.

`fledge ai status` does show `(from env)` which is helpful, but:
1. `fledge config list` doesn't indicate that the displayed host value is overridden by an env var
2. When `fledge ask` fails with a connection error, the error message doesn't mention the env var override

### Steps to reproduce

```bash
export OLLAMA_HOST=localhost:11434

# Set a different host in config
fledge config set ai.ollama.host "https://cloud.ollama.com"

# Config list shows the config value but env var will override it
fledge config list
# Shows: ai.ollama.host  https://cloud.ollama.com

# But ai status reveals the truth
fledge ai status
# Shows: Host: http://localhost:11434 (from env)

# ask uses the env var, not the config
fledge ask "hello"
# error: POST http://localhost:11434/api/generate ...
```

### Expected behavior

- `config list` should annotate overridden values: `ai.ollama.host  https://cloud.ollama.com  (⚠️ overridden by OLLAMA_HOST env var)`
- Error messages from `ask` should mention when env vars are overriding config: `"Connecting to http://localhost:11434 (from OLLAMA_HOST env var — unset it to use config value)"`

### Impact

Users set the config, think it should work, but the env var silently wins. This is especially confusing when combined with the `config unset` bug where the default value re-appears.

### Environment

- fledge 1.3.1
- macOS (Darwin 25.3.0)

---
🤖 Agent: CorvidAgent | Model: Opus 4.6
👤 Requested by: @Kyntrin

0xLeif:
Fixed in #384.

PR #384, opened 2026-05-12
fix: address open issues #377, #378, #379, #382, #383

### LINKED ISSUE #377: `config unset ai.ollama.host` silently restores default instead of clearing
## Bug

`fledge config unset ai.ollama.host` reports `✅ Unset ai.ollama.host` but the value immediately snaps back to the hardcoded default `http://localhost:11434` in the config file.

### Steps to reproduce

```bash
# Change host to something custom
fledge config set ai.ollama.host "https://custom.endpoint.com"
# Verify it saved
cat "$(fledge config path)"  # shows host = "https://custom.endpoint.com" ✅

# Now try to unset it
fledge config unset ai.ollama.host
# Reports: ✅ Unset ai.ollama.host

# But check the file again
cat "$(fledge config path)"  # shows host = "http://localhost:11434" ← default restored!
```

### Expected behavior

Either:
1. The field should actually be removed from the TOML file so the runtime default is used without persisting it, OR
2. `unset` should warn that this field has a hardcoded default: `"ai.ollama.host has a default value (http://localhost:11434) — unset will restore the default, not remove the field"`

### Actual behavior

`unset` says success, the field is momentarily cleared, but the config serialization immediately re-applies the `#[serde(default)]` value and writes it back to disk. The user has no way to distinguish "explicitly set to localhost" from "using the default."

### Impact

Users trying to switch from local Ollama to Ollama Cloud cannot remove the host field — they must explicitly `set` it to their cloud URL instead. The `unset` command is misleading.

### Environment

- fledge 1.3.1
- macOS (Darwin 25.3.0)

---
🤖 Agent: CorvidAgent | Model: Opus 4.6
👤 Requested by: @Kyntrin

0xLeif:
## Summary

Closes #377, #378, #379, #382, #383.

| # | Fix |
|---|---|
| **#383** | `spec new game/board` now creates `specs/game/board/board.spec.md` with `module: game/board` in frontmatter and a quoted nested key in `registry.toml`. Validator still rejects `\`, leading/trailing `/`, `//`, and any `..`/`.` segment. |
| **#382** | `plugins update` skips entries without a `.git` (dir or file — covers worktrees/submodules) with an info message rather than warning on a doomed `git pull`. Matches workspace-managed plugin layouts like Merlin. |
| **#379** | New `ai.claude.api_key` config key (secret). `build_provider` forwards it to the `claude` CLI via `ANTHROPIC_API_KEY`. Env var wins. `config list` and `ai status` surface the source. |
| **#378** | Ollama connection errors and HTTP errors now append `(OLLAMA_HOST env var = …; unset it to use ai.ollama.host config)` when the env var is set, so users stop guessing why the override won. |
| **#377** | `OllamaConfig.host`/`model`/`timeout_seconds` get `skip_serializing_if` keyed on the runtime default, so `config unset` truly removes the field from the TOML instead of re-persisting `http://localhost:11434`. |

## Test plan

- [x] `fledge run test` — 1032 passed, 0 failed
- [x] `fledge run lint` — clean
- [x] `fledge run fmt` — clean
- [x] `fledge run spec-check` — 30 specs, 0 errors / 0 warnings
- [x] Manual smoke against the rebuilt debug binary:
  - `spec new game/board` produces the expected nested layout + frontmatter + quoted registry key
  - `spec new` rejects `../evil`, `foo//bar`, `/foo`, `foo\bar`, `foo/../bar`
  - `config set` + `config unset ai.ollama.host` leaves `[ai.ollama]` empty in the TOML
  - `config list` annotates `(⚠ overridden by OLLAMA_HOST env)` and `(from ANTHROPIC_API_KEY env)` when those env vars are set
  - `ask` connection error includes the `OLLAMA_HOST env var = …; unset it…` hint
  - `ai status` reports the Claude API key source (`from env` / `from config` / not-set hint)
  - `is_git_repo` unit tests cover `.git` dir, `.git` file, and workspace (no `.git`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ Cargo.lock
 docs/book/
 .claude/
 .reviews/
+.fledge/
--- a/specs/config/config.spec.md
+++ b/specs/config/config.spec.md
@@ -1,6 +1,6 @@
 ---
 module: config
-version: 11
+version: 12
 status: active
 files:
   - src/config.rs
@@ -26,8 +26,8 @@ Manages global user configuration from `~/.config/fledge/config.toml`. Provides 
 | `TemplatesConfig` | Struct holding additional template directory paths and remote repo references |
 | `GitHubConfig` | Struct holding optional GitHub token for authenticated access |
 | `AiConfig` | Struct holding LLM provider selection and per-provider settings |
-| `ClaudeConfig` | `{ model: Option<String> }` — optional default model passed to `claude --model` |
-| `OllamaConfig` | `{ host: String, api_key: Option<String>, model: String, timeout_seconds: u64 }` — endpoint, auth, default model, and per-request timeout |
+| `ClaudeConfig` | `{ model: Option<String>, api_key: Option<String> }` — optional default model passed to `claude --model`, plus optional Anthropic API key (falls back to `ANTHROPIC_API_KEY` env var) |
+| `OllamaConfig` | `{ host: String, api_key: Option<String>, model: String, timeout_seconds: u64 }` — endpoint, auth, default model, and per-request timeout. `host`/`model`/`timeout_seconds` are `skip_serializing_if` default so `config unset` truly removes them from the TOML |
 | `TrustConfig` | `{ orgs: Vec<String>, users: Vec<String> }` — additional GitHub orgs and users to trust at team tier |
 | `load` | Loads config from disk or returns defaults if file is missing |
 | `config_path` | Returns the platform-appropriate path to the config file |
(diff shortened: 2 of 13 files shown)


corvid-agent:

corvid-agent (dismissed):
## Review — CorvidAgent (Rook)

Clean, well-structured PR. All five fixes are correct and well-tested. Approving.

### Per-issue notes

**#377 — `config unset` re-persists defaults**
`skip_serializing_if` with predicate functions against the default values is the right approach. The round-trip test (`unset_ollama_host_does_not_persist_default`) is thorough — verifies host, model, and timeout in one go.

**#378 — Ollama env var hint**
`ollama_host_env_hint()` returning an empty string when unset is clean — lets call sites append unconditionally without branching. Good ergonomics.

**#379 — Claude API key config**
Properly marked as a secret key, env var correctly takes precedence over config, and the `config list` / `ai status` display logic mirrors the existing Ollama pattern. One minor note: `build_claude_env_beats_config_api_key` doesn't actually assert precedence (can't read `api_key` off the trait object) — it just verifies the build succeeds. The real precedence verification lives in `resolve_claude_api_key_source`, which is fine, but the test name slightly oversells what it checks.

**#382 — `plugins update` skips workspace-managed plugins**
`is_git_repo` checking `.git` existence (covering both dirs and files for worktrees/submodules) is correct and minimal. Good test coverage with the three cases (dir, file, absent).

**#383 — Nested spec names**
The `module_leaf()` helper and per-segment validation in `validate_module_name` are well done. Quoting nested keys in `registry.toml` is a nice detail. The updated `test_load_module_bundle_rejects_path_traversal` correctly drops `foo/bar` from the rejection list with a clear comment explaining why.

### Nit (non-blocking)

In `module_leaf`, `rsplit('/').next()` always returns `Some` for a non-empty iterator, so the `.unwrap_or(name)` fallback is dead code. Harmless but could be `.unwrap()` or just documented. Not worth a re-spin.

LGTM — ship it. 🚢

PR #386, opened 2026-05-17
🧹 [code health improvement] Remove unused schema_version field from FledgeFile

0xLeif:
🎯 **What:** Removed the unused `schema_version` field from the `FledgeFile` struct in `src/run.rs`.
💡 **Why:** This is dead code that was marked with `#[allow(dead_code)]`. Removing it improves code health and readability without changing behavior.
✅ **Verification:** Verified by running `cargo test` which confirmed no regressions or broken functionality. 
✨ **Result:** A cleaner struct definition that only contains fields that are actively used.

---
*PR created automatically by Jules for task [3929878372658690201](https://jules.google.com/task/3929878372658690201) started by @0xLeif*

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -16,8 +16,6 @@ const RUN_INIT_SCHEMA: u32 = 1;
 struct FledgeFile {
     #[serde(default)]
     tasks: BTreeMap<String, TaskDef>,
-    #[allow(dead_code)]
-    schema_version: Option<u32>,
 }
 
 #[derive(Debug, Deserialize)]


corvid-agent (approved):
Approving the field removal — `#[allow(dead_code)]` confirmed it was never read, and serde's default behavior silently ignores unknown TOML fields, so existing `fledge.toml` files with a stray `schema_version` entry won't break.

Minor follow-up (non-blocking): the parenthetical in the `RUN_*_SCHEMA` doc comment at src/run.rs:8-10 — 'distinct from the `schema_version` field on `fledge.toml` itself, which is a manifest version' — references the field this PR removes and becomes slightly orphaned. Worth a one-line cleanup in a follow-up.

PR #394, opened 2026-05-17
🔒 Fix Command/Flag Injection in git checkout (plugin install and update)

0xLeif:
🎯 **What:** Command/Flag Injection in `git checkout` during plugin installation and update.
⚠️ **Risk:** A malicious or malformed plugin source could provide a git reference starting with a hyphen (e.g., `-B`), causing `git checkout` to interpret it as a command-line option rather than a reference, potentially leading to unauthorized execution flags or failures.
🛡️ **Solution:** Added explicit validation to check if the git reference (`ref_str` or `tag`) starts with a hyphen (`-`). If it does, the operation fails and the installation is aborted with an appropriate error message, cleaning up any partially installed files. This ensures that user-provided data cannot inject flags into the git command.

---
*PR created automatically by Jules for task [1711700031451668183](https://jules.google.com/task/1711700031451668183) started by @0xLeif*

### DIFF
--- a/src/plugin/install.rs
+++ b/src/plugin/install.rs
@@ -260,6 +260,13 @@ pub(crate) fn install_plugin(source: &str, force: bool, json: bool) -> Result<se
     }
 
     if let Some(ref_str) = git_ref {
+        if ref_str.starts_with('-') {
+            fs::remove_dir_all(&plugin_dir).ok();
+            bail!(
+                "Invalid git ref '{}': references cannot start with a hyphen.",
+                ref_str
+            );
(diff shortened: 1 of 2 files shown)


corvid-agent (approved):
Approving. Defensive validation against argv-prefix flag injection on `git checkout` is the right call here. Rejecting refs starting with `-` is the standard mitigation (git doesn't support `--` as an end-of-options separator for checkout's ref arg). install.rs also cleans up the partial clone on failure. Both call sites covered.

PR #397, opened 2026-05-17
🔒 Fix command injection vulnerability in git diff

0xLeif:
🎯 **What:** The vulnerability fixed
A Command/Flag Injection vulnerability in `src/review.rs` caused by passing an unsanitized `base` parameter (derived from CLI input) into `git diff`, which allowed git to interpret inputs starting with a hyphen as arbitrary flags.

⚠️ **Risk:** The potential impact if left unfixed
An attacker (or user tricked into passing a malicious string) could pass arbitrary flags to `git diff` (like `--output=path`), leading to arbitrary file overwrite and potentially arbitrary code execution if critical files are overwritten.

🛡️ **Solution:** How the fix addresses the vulnerability
This change adds an early validation check in `get_diff`, `get_diff_stats`, and `get_changed_files` to ensure that `base` does not start with a hyphen (`-`). If it does, the function immediately bails out. Because valid Git references and commit hashes cannot start with a hyphen, this prevents the flag injection without affecting valid use cases.

---
*PR created automatically by Jules for task [11650463467420088031](https://jules.google.com/task/11650463467420088031) started by @0xLeif*

### DIFF
--- a/src/review.rs
+++ b/src/review.rs
@@ -487,6 +487,9 @@ fn default_branch() -> Result<String> {
 }
 
 fn get_diff(base: &str, file: Option<&str>) -> Result<String> {
+    if base.starts_with('-') {
+        bail!("Invalid base revision: cannot start with '-'");
+    }
     let mut args = vec!["diff", base];
     if let Some(f) = file {
         args.push("--");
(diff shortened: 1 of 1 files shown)


corvid-agent (approved):
Approving. Defensive guard added in all three call sites (`get_diff`, `get_diff_stats`, `get_changed_files`). Valid git refs never start with a hyphen, so this is a non-breaking mitigation against `--upload-pack`-style flag injection.

PR #410, opened 2026-06-03
Support local plugin installs

0xLeif:
## Summary
- Add local path plugin installs with live symlink mode and --copy snapshots
- Support generic git URL plugin sources while preserving GitHub shorthand
- Add local trust tier, safe local removal, update skipping, tests, specs, and introspection snapshot updates

## Test Plan
- [x] fledge run test --json
- [x] fledge run spec-check --json
- [x] fledge lanes run pre-commit --json


0xGaspar (approved):
LGTM

PR #424, opened 2026-06-23
Delegate `fledge spec check` to the real specsync binary for CI parity

0xLeif:
## Summary

`fledge spec check` only validated spec **structure** (frontmatter, required sections, files exist). CI runs `CorvidLabs/spec-sync`, which additionally parses each governed source file's **public exports** and fails when they drift from the spec's Public API tables. The two checkers were orthogonal — a spec could pass `fledge spec check` locally yet fail spec-sync in CI (the green-locally/red-in-CI trap).

This closes the gap **by construction**: when the `specsync` binary is on `PATH`, `spec check` shells out to it (the same binary CI runs), so a local pass guarantees a CI pass. When absent, it falls back to today's structural check with an install hint.

Rather than re-implement spec-sync's ~3.7k-LOC, per-language export-AST layer inside fledge (which would just create a second checker that drifts from upstream), we delegate to the real, separately-versioned tool.

## Changes

- **`src/spec/engine.rs`** *(new)* — `find_specsync` (PATH lookup) + `try_check_via_specsync`. Invokes `specsync check --force --format json --root <root>` (+`--strict`), parses the `{passed, errors, warnings, stale, specs_checked}` report, renders fledge-style. `--force` mirrors the CI action so results match a fresh run, not the local hash cache.
- **`src/spec/commands.rs`** — delegation runs first; structural fallback + `cargo install specsync` hint when absent. JSON output gains an `engine` field (`"specsync"` | `"structural"`).
- **`src/spec/mod.rs`** — registers the module.
- **`specs/spec/spec.spec.md`** → **v12**, documenting the new `engine` module.

## Behavior

| State | `fledge spec check` |
|-------|---------------------|
| `specsync` on PATH | Delegates → identical to CI (export-coverage validation) |
| `specsync` absent | Structural check (as before) + install hint |

Note: delegation runs whatever `specsync` is installed. CI currently pins `v4.3.2`; for byte-identical parity, contributors should install the pinned version.

## Test Plan

- [x] Built `specsync` v4.5.0 from source; `fledge spec check --strict` through it reports **0 errors / 0 warnings / exit 0** against this repo (engine=specsync)
- [x] Stub-binary tests confirm exact arg forwarding (`check --force --format json --root … [--strict]`), non-zero exit on findings, both render branches
- [x] Fallback path (no specsync) unchanged; emits install hint
- [x] New unit tests for report parsing (full + minimal) and rendering
- [x] `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt --check` all pass

> Dogfooding catch: the real specsync immediately flagged that adding `pub(crate) mod engine` introduced an undocumented export that **would have failed CI under strict** — exactly the trap this PR eliminates. It's documented in `spec.spec.md` v12, which is why the check is clean.

### DIFF
--- a/specs/spec/spec.spec.md
+++ b/specs/spec/spec.spec.md
@@ -1,12 +1,13 @@
 ---
 module: spec
-version: 11
+version: 12
 status: active
 files:
   - src/spec/mod.rs
   - src/spec/parse.rs
   - src/spec/validation.rs
   - src/spec/commands.rs
+  - src/spec/engine.rs
   - src/spec/tests.rs
 
 db_tables: []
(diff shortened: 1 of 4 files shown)


Kyntrin (approved):
I trust Leif, LGTHIM!

PR #434, opened 2026-07-02
Fix: reject dot-dot path segments in plugin sources — trust-tier spoof (H-2)

0xLeif:
## Summary

The trust classifier keyed on the **first** path segment of a source, so `CorvidLabs/../attacker/evil` classified as the official CorvidLabs org (green "official CorvidLabs plugin" banner) — while git/curl collapse the clone URL client-side (RFC 3986 dot-segment removal) to `attacker/evil`. Since only the Unverified tier is capability-blocked, the attacker repo's `exec`/`network` capabilities were allowed (auto-granted under `--force` / non-interactive). Verified live: `git ls-remote` of `.../CorvidLabs/../attacker/evil.git` fetches `attacker/evil`.

- `trust.rs`: `classify_source` returns `Unverified` for any source with a `.`/`..` segment; adds `source_has_path_traversal()`.
- `install.rs`: `InstallSource::parse` rejects such sources outright (before classification or clone), shorthand and full-URL forms alike.
- Tests: `dotdot_source_never_official`, `source_traversal_detection`, `parse_rejects_dotdot_source`, `parse_accepts_normal_shorthand`.

Addresses review finding **H-2**.

## Test Plan

- [ ] New tests pass; `cargo fmt --check` + `cargo clippy -- -D warnings` clean
- [ ] Full unit suite green (876 passed locally)

### DIFF
--- a/specs/trust/trust.spec.md
+++ b/specs/trust/trust.spec.md
@@ -1,6 +1,6 @@
 ---
 module: trust
-version: 4
+version: 5
 status: active
 files:
   - src/trust.rs
@@ -28,6 +28,7 @@ Shared trust-tier classification for all extension types (plugins, templates, la
 | `determine_trust_tier` | Classify a source string into a trust tier by parsing the GitHub org/owner |
 | `determine_trust_tier_from_owner` | Classify by GitHub owner string directly |
 | `parse_source_ref` | Split a source string into base URL and optional git ref (e.g., `@v1.0.0`) |
+| `source_has_path_traversal` | Detect a `.` or `..` segment in a source's owner/repo path (used to reject trust-tier spoofing via clone-URL collapse) |
 
 ### Structs & Enums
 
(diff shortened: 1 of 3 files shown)


0xLeif:
Addressed in b50617a. Gemini's HIGH finding was correct — the `.`/`..` segment check only inspected the raw source string, so percent-encoded traversal (`%2e%2e`, `%2f`, `%5c`) slipped past and `github.com/Official/%2e%2e/evil` still classified as Official. `path_has_traversal_segment` now checks both the raw source **and** a single-pass percent-decoded copy (matching git/curl transport decoding), with tests for lower/upper-case `%2e%2e`, `%2f`, `%5c`, a decoded `determine_trust_tier` spoof, and a literal `%25` non-false-positive. Thanks for the catch.

Kyntrin (approved):
Reviewed the current head. The path traversal fix covers literal and percent-encoded dot/slash/backslash cases, rejects bad install sources before clone, and keeps normal shorthand parsing intact.

Verified locally:
- cargo test trust::tests:: -- --nocapture
- cargo test plugin::install::tests:: -- --nocapture
- FLEDGE_NON_INTERACTIVE=1 fledge spec check --json
- cargo fmt --check

PR #437, opened 2026-07-02
Fix: prevent silent capability escalation on plugin update (H-3)

0xLeif:
## Summary

`fledge plugins update` only diffed `exec`/`store`/`metadata` when deciding whether a freshly-fetched manifest requested new capabilities — `network` and `filesystem` were never compared. So an installed plugin could push a commit adding `network = true` (the WASM exfiltration channel) or escalating `filesystem` from `none` → `project`/`plugin`, and `fledge plugins update` granted it **silently, with no prompt** (even non-interactively). The update path also never re-ran the trust-tier gate, so an **unverified** plugin could gain `exec`/`network` via an update that a fresh install would refuse outright. Contradicts SECURITY.md ("all capabilities are prompted at install time"; "network/filesystem … require user approval").

- Extracts `CapabilityDelta` (pure, unit-tested) that diffs **network and filesystem** too, treating any move to a non-`none` filesystem mode as an escalation.
- Re-runs `check_tier_capabilities` against the source's trust tier on the newly-added `exec`/`network` before persisting — unverified sources are refused with a clear `failed` status (keeping `update --all`/`--defaults` fault-tolerant).
- Prompts for newly-added `network`/`filesystem` alongside `exec`/`store`/`metadata`. Unchanged-cap updates are unaffected (no prompt, no gate).

## Test Plan

- [x] 11 new unit tests: added network/filesystem, project→plugin escalation, downgrade ignored, no-op, tier gating (exec/network blocked for unverified, allowed for official), filesystem not tier-gated, wasm `None`-baseline fail-closed, unchanged dangerous caps don't retroactively block
- [x] Full unit suite green (883 passed); `clippy -D warnings` clean; `fmt --check` clean; `spec check` clean

## Note

Designed and adversarially reviewed before implementation. That review caught a compile blocker — `check_tier_capabilities` was `#[cfg(test)]`-gated in `plugin/mod.rs` — which this PR fixes by making that import unconditional (it is now used by the update path, not only tests).

### DIFF
--- a/src/plugin/mod.rs
+++ b/src/plugin/mod.rs
@@ -35,7 +35,6 @@ use search::search_plugins;
 use update::update_plugins;
 use validate::validate_plugin;
 
-#[cfg(test)]
 use install::check_tier_capabilities;
 #[cfg(test)]
 use run_plugin::{apply_protocol, resolve_plugin_source_dir, which_fledge_plugin};
--- a/src/plugin/update.rs
+++ b/src/plugin/update.rs
@@ -357,12 +357,24 @@ struct CapabilityDelta {
 
 impl CapabilityDelta {
     fn compute(old: Option<&PluginCapabilities>, new: &PluginCapabilities) -> Self {
-        // Filesystem is an escalation ladder (none → project → plugin/custom);
-        // any change to a non-"none" mode grants new host access, so treat it
-        // as newly added. A missing entry (`None`) is equivalent to "none".
+        // Filesystem is an escalation ladder: none < project < plugin. Only a
+        // move UP the ladder grants new host access; a downgrade (e.g.
+        // plugin → project, which drops plugin-data write while keeping project
+        // read) or a same-level change is not an escalation and must not
+        // re-prompt. An unrecognized mode is treated as the most-privileged
+        // rung (fail-closed) so a novel mode always re-prompts. A missing entry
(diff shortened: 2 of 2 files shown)


Kyntrin (changes requested):
One issue to fix before merge: the filesystem capability diff treats every non-none mode change as a new grant.

That catches real escalations like none -> project and project -> plugin, but it also flags plugin -> project. Since plugin mode already includes project read access plus plugin data write access, plugin -> project is a downgrade. Today that downgrade would still prompt, and would fail non-interactive updates.

Please compare filesystem modes by privilege level and add a regression test for plugin -> project.

0xLeif:
Addressed in e0d0d57 — two fixes from review:

1. **@Kyntrin / Gemini** — filesystem escalation was "any change to a non-`none` mode", which wrongly flagged the `plugin → project` *downgrade* (drops plugin-data write while keeping project read). Now ranked (`none < project < plugin`; an unknown mode = most-privileged, so fail-closed) and only a move **up** the ladder re-prompts. Regression test `delta_ignores_filesystem_downgrade_plugin_to_project` added.
2. **Gemini (HIGH)** — the `--json` + interactive grant prompt showed no capability list (it was gated behind `if !json`). The list is now built once and routed to **stderr** in `--json` mode (stdout stays a clean JSON document) or stdout otherwise, so an operator always sees what they are granting before the confirm.

Thanks both.

PR #440, opened 2026-07-02
Fix: atomic write for config.toml — no torn write of the API-key file (Medium)

0xLeif:
## Summary

`Config::save()` open-truncate-wrote `~/.config/fledge/config.toml` **in place** — the same read-modify-write / torn-read race class [#419](https://github.com/CorvidLabs/fledge/pull/419) fixed for `plugins.toml`, except this is the file holding `github.token` and every `ai.*.api_key`. A crash or concurrent reader mid-write could corrupt or lose secrets.

`save()` now delegates to a new `save_to()`, which:
- writes to a **same-directory** temp file (`<dir>/config.toml.tmp.<pid>`) so the rename is atomic on the same filesystem,
- `create_dir_all`s the parent first (config dir may not exist),
- atomically renames it over the target,
- sets `0600` perms on unix (it holds secrets).

Mirrors the existing `save_registry_to` pattern.

## Test Plan

- [x] 3 new unit tests: writes secrets + leaves no temp file behind; fully replaces an existing file (no stale bytes); sets owner-only `0600` on unix
- [x] full `config::` suite green (64 passed); `clippy -D warnings` + `fmt --check` clean

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -235,38 +235,48 @@ impl Config {
     /// corrupt or lose the file, and this file holds API keys — the same
     /// torn-write class PR #419 fixed for plugins.toml. Same-directory rename
     /// is atomic on POSIX, so readers see either the old config or the new one.
-    /// On unix the temp file is created (and pinned) 0600 since it holds
-    /// secrets, and the rename carries those perms onto the target.
+    /// On unix the temp file is created 0600 (owner-only) via an exclusive
+    /// create, since it holds secrets, and the rename carries those perms onto
+    /// the target.
     fn save_to(&self, path: &Path) -> Result<()> {
         if let Some(parent) = path.parent() {
             std::fs::create_dir_all(parent)?;
         }
         let content = toml::to_string_pretty(self)?;
-        let tmp = path.with_extension(format!("toml.tmp.{}", std::process::id()));
+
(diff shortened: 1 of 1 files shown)


Kyntrin (changes requested):
Requesting changes for one blocking issue.

Medium: `src/config.rs` uses a predictable temp path, `config.toml.tmp.<pid>`, and opens it with `create(true).truncate(true)`. Because this file stores API keys, a stale/reused temp file can keep permissive mode until after secrets are written, and concurrent same-process saves can collide on the same temp path. Please use a unique temp name plus exclusive creation, or a same-directory temp-file helper that guarantees exclusive creation, and set permissions on the open file handle before writing.

Verification I ran:
- `cargo test config::tests::save_to_` passed.
- `cargo test config::` passed: 64 tests passed.

The current tests do not cover stale temp files or concurrent saves.

0xLeif:
Addressed in 7e049cf. Switched to exclusive `create_new(true)` + `mode(0o600)` so the secrets file is owner-only **from birth** (no loose-permission window, and no path-based `set_permissions` TOCTTOU — dropped it), and added a per-process atomic counter to the temp name so concurrent same-process saves can't collide on the path. Thanks @Kyntrin.

PR #441, opened 2026-07-02
Fix: stop swallowing git errors in templates init + publish push (Medium)

0xLeif:
## Summary

Two error-swallowing bugs from the review:

- **`templates init` git status** — `init_git` ran `git add`/`git commit` with `.status()` and stderr nulled, and both call sites hardcoded `git_initialized: !opts.no_git`. So a failed `add`/`commit` (failing hook, missing identity, …) still reported `git_initialized: true` at exit 0. `init_git` now returns whether the commit actually succeeded — treating `"nothing to commit"` as benign so an empty tree isn't a false failure — warns to stderr on real failures, and the JSON envelope reports the true state.
- **publish push stderr** — the git push piped stderr but called `.status()`, never reading it: git's error detail was lost and a full stderr buffer could deadlock the child. Now uses `.output()` and surfaces the error on failure.

## Test Plan

- [x] New regression test: a failing `pre-commit` hook makes `init_git` return `false` (unix-gated); existing success-path tests still green (now `Ok(true)`)
- [x] Full unit suite green (891 passed); `clippy -D warnings` + `fmt --check` clean

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -145,9 +145,11 @@ pub fn run(mut opts: InitOptions) -> Result<()> {
     )?;
 
     // Git init
-    if !opts.no_git {
-        init_git(&target_dir)?;
-    }
+    let git_initialized = if opts.no_git {
+        false
+    } else {
+        init_git(&target_dir)?
+    };
 
     let hooks_run = !opts.no_install && reqs_ok && !template.manifest.hooks.post_create.is_empty();
     // Post-create hooks for **local** templates: `--yes` is sufficient consent
@@ -172,7 +174,7 @@ pub fn run(mut opts: InitOptions) -> Result<()> {
             None,
             &variables,
             &created_files,
-            !opts.no_git,
+            git_initialized,
             hooks_run,
         )?;
     } else {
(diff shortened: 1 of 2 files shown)


Kyntrin (changes requested):
Requesting changes for two issues.

High: `src/publish.rs` now includes raw `git push` stderr in the user-facing error. This push injects the GitHub token through `http.extraheader`, and this repo already treats that stderr as a redaction boundary in `remote.rs` using `crate::utils::redact_secrets`. Please redact `stderr` before adding it to the `bail!` message.

Medium: `src/init.rs` updates the JSON envelope to use the true `git_initialized` value, but the non-JSON summary still calls `print_summary(..., opts.no_git)`. That means it still prints “git repo initialized with initial commit” whenever `--no-git` was not passed, even if `init_git()` returned `false`. Please pass the actual `git_initialized` result through the summary path too.

Verification I ran:
- `cargo test init_git` passed: 3 tests passed.
- `cargo test publish` passed: 3 passed, 1 ignored.

0xLeif:
Addressed in 61c28cb — all three:
1. **HIGH (token leak):** the push stderr is now run through `crate::utils::redact_secrets` before being surfaced in the error, matching the redaction boundary in `remote.rs`.
2. Replaced the locale-fragile `"nothing to commit"` string check with `git diff --cached --quiet` (skip commit when nothing is staged; any commit failure with staged changes is a real failure).
3. `print_summary` now receives the true `git_initialized` value, so the non-JSON summary no longer claims a git repo was initialized when the commit failed.

Thanks @Kyntrin.

PR #442, opened 2026-07-02
Update: fix stale provider --help text + SECURITY.md accuracy (Medium)

0xLeif:
## Summary

Doc-accuracy drift caught by the review (companion to the H-5 spec fixes):

- **`src/cli.rs` help text** — 8 `--help` doc comments still described the pre-1.5.0 model: "LLM provider: claude (default) or ollama" and "ai.{claude,ollama}.model". The default is now **auto-detect** and the provider set is broader. Reworded to match. The `value_parser` arrays are unchanged — `claude` is still accepted as a **deprecated alias**, so it stays in the list.
- **SECURITY.md** — Supported Versions was stuck at `1.0.x` with "Once 1.0 ships…" prose while the project is at **1.6.0** (updated to `1.6.x`). Separately, the "prompt/cancel timeouts at 5 minutes" line was inaccurate — there is no prompt/cancel timeout in the code; the real 5-minute limit is `MAX_EXEC_TIMEOUT` on **exec** commands (default 30s, capped at 300s), so I restated it accurately (verified against `src/protocol/exec.rs`).

## Test Plan

- [x] `introspect` JSON snapshot updated — diff is exactly the 8 help-text edits, no structural change
- [x] Full unit suite green (890 passed); `clippy -D warnings` + `fmt --check` clean

## Follow-up (not in this PR)

Docs hardcode `~/.config/fledge/config.toml` in ~8 places; on macOS/Windows the code uses the platform config dir. That needs per-OS wording across the docs site and is better as its own change — filing as an issue.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,12 +4,11 @@
 
 | Version | Supported |
 |---------|-----------|
-| 1.0.x   | Yes              |
-| 0.17.x  | Best effort      |
-| < 0.17  | No               |
+| 1.6.x   | Yes              |
+| < 1.6   | Best effort      |
 
-Once 1.0 ships, only 1.0.x receives security fixes. Until then, the latest
-0.x release is the supported line.
+The latest 1.x release is the supported line; earlier releases receive
+best-effort fixes only.
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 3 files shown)


Kyntrin (changes requested):
Requesting changes for help-text accuracy.

Medium: `src/cli.rs` says `ai use` provider defaults to the active provider, but the implementation does not do that. With no provider, `use_provider()` calls `require_interactive("provider")` and prompts; in non-interactive mode it errors. Please use wording like `Provider (interactive prompt if omitted)`.

Low: the `--with-model anthropic` example says it uses the active provider config, but it resolves that provider’s specific config (`ai.anthropic.model`, then legacy `ai.claude.model`), not the globally active provider. Wording like `that provider’s config` or `active anthropic config` would be clearer.

Verification I ran:
- `cargo test introspect` passed.
- `cargo test ai::` passed.
- `cargo test cli_ai` passed.

0xLeif:
Good catches — both were inaccuracies I introduced, fixed in the latest commit:
- `ai use` provider help now reads **"(interactive prompt if omitted)"** — it prompts (or errors non-interactively), it does not default to the active provider.
- `--with-model anthropic` help now says **"active anthropic config"** (provider-specific `ai.anthropic.*`), not "active provider config". Thanks @Kyntrin.

PR #456, opened 2026-07-02
Fix: retry formula-convergence check for Contents API propagation lag

0xLeif:
## Summary

The "Verify channels converged" step (added in #435) read the just-pushed Homebrew-tap formula via the GitHub **Contents API** and asserted the version immediately. That API serves a **cached blob for several seconds** after a push, so a sync that *actually landed* false-failed as "formula sync did not take effect".

Observed live on the **v1.6.0 backfill**: the sync step pushed `chore: update formula to v1.6.0` and the tap formula is now `1.6.0`, yet the verify step read a stale `1.4.3` and exited 1 — a red run for a successful sync.

Fix: poll the Contents API up to **6× with a 10s backoff** before declaring failure. A genuine no-op push still goes red; only the propagation race is removed.

## Test Plan

- [x] YAML validates (`yaml.safe_load`)
- [x] Verified against the real failure: the tap **did** converge to 1.6.0 (commit `chore: update formula to v1.6.0`, 14:33); only the eager read failed
- [ ] Confirmed green on the next release / `workflow_dispatch` backfill

### DIFF
--- a/.github/workflows/post-release-formula.yml
+++ b/.github/workflows/post-release-formula.yml
@@ -138,13 +138,22 @@ jobs:
           NEW_VERSION: ${{ steps.tag.outputs.version }}
         run: |
           set -euo pipefail
-          # Read the just-pushed formula via the Contents API (not the CDN, which
-          # can lag) and assert the version actually took effect. This turns a
-          # silent no-op push into a red run instead of leaving brew users stale.
-          formula="$(gh api repos/CorvidLabs/homebrew-tap/contents/Formula/fledge.rb --jq '.content' | base64 --decode)"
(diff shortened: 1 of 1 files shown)


0xGaspar (approved):
Correct fix for the propagation race — the diagnosis (Contents API serving a cached blob for seconds post-push) matches the v1.6.0 backfill evidence, and polling 6×/10s while still going red on a genuine no-op preserves the original guard's intent. Early `break` on match keeps the happy path fast. Two minor notes: under `set -euo pipefail`, a transient `gh api` failure inside the loop would abort the script rather than retry (the loop only absorbs propagation lag, not API errors); and on a real failure the 6th attempt still sleeps 10s before the final check. Neither is blocking. LGTM.

PR #459, opened 2026-07-02
Update: enforce Cargo.lock in CI with --locked

### LINKED ISSUE #450: CI: enforce the committed lockfile with --locked
Now that `Cargo.lock` is committed (#436), CI's build and `cargo audit` steps still resolve dependencies fresh (`cargo generate-lockfile`) instead of honoring the lock. Switch build/test/audit to `--locked` so the committed lockfile is enforced and CI reflects the exact shipped dependency set. Otherwise the lockfile is tracked but not authoritative.

---
_Filed from a skeptical staff-level review (Medium-severity tier). All Critical/High findings and the correctness/doc Medium findings were fixed in #432–442._

0xLeif:
## Summary
- Now that `Cargo.lock` is committed (#436), make CI enforce it by adding `--locked` to the cargo build/test invocations.
- `.github/workflows/ci.yml`: `cargo test --verbose --locked` (test job) and `cargo build --release --locked` (integration job).
- `.github/workflows/release.yml`: `cargo test --verbose --locked` (test gate) and `cargo build --release --target ... --locked` (both cross and native build steps).
- Deliberately left untouched: the `audit` job (rustsec/audit-check manages its own lockfile via `cargo generate-lockfile`) and `cargo fmt`/`cargo clippy` steps.

Closes #450

## Test Plan
- [x] YAML validated as well-formed (`yaml.safe_load` on both files).
- [x] Diff is minimal: 5 insertions, 5 deletions across the two workflow files.
- [ ] CI runs `--locked` on the PR; a stale `Cargo.lock` would now fail the build/test steps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@v5
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
-      - run: cargo test --verbose
+      - run: cargo test --verbose --locked
 
   lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


0xGaspar (approved):
Right call now that `Cargo.lock` is committed — `--locked` makes a stale lockfile fail loudly. Good judgment leaving `audit` alone (rustsec regenerates its own lockfile) and skipping fmt/clippy where it's a no-op. LGTM.

PR #461, opened 2026-07-02
Update: make wasmtime deps optional behind a `wasm` feature (on by default)

### LINKED ISSUE #445: Build: feature-gate the WASM plugin runtime
`wasmtime` + `wasmtime-wasi` are unconditional dependencies, pulling ~36 wasm/cranelift crates into every build for an optional plugin runtime. Gate the wasm runtime behind a Cargo feature so users who don't run wasm plugins get faster builds and a smaller binary. (Note: keep it default-on if wasm plugins are considered core.)

Files: `Cargo.toml`, `src/plugin/wasm.rs`.

---
_Filed from a skeptical staff-level review (Medium-severity tier). All Critical/High findings and the correctness/doc Medium findings were fixed in #432–442._

0xLeif:
## Summary
- `wasmtime` and `wasmtime-wasi` were unconditional dependencies (~36 transitive crates) purely for the optional sandboxed WASM plugin runtime. Marked both `optional = true` and added a `[features]` block with `default = ["wasm"]` and `wasm = ["dep:wasmtime", "dep:wasmtime-wasi"]`. Default behavior is unchanged; `--no-default-features` drops the entire wasmtime dependency tree.
- Cfg-gated every wasm code path on `feature = "wasm"`:
  - `mod wasm;` in `src/plugin/mod.rs`
  - the WASM run-time dispatch in `src/plugin/run_plugin.rs`
  - the WASM pre-compile step in `src/plugin/install.rs`
- For the disabled path, both the run and install sites now `bail!` with a clear error: `"Plugin '<name>' requires the WASM runtime, which was not compiled in (rebuild with --features wasm)."`
- Gated the `protocol` re-exports (`detect_project_context`, `handle_exec`, `handle_metadata`, `handle_load`, `handle_store`, `handle_log`) that exist solely for the wasm host bridge, so the no-default-features build stays warning-free (they are the only external consumers).

## Test Plan
- [x] `cargo build` (default, wasm ON) — clean
- [x] `cargo build --no-default-features` — clean, warning-free
- [x] `cargo tree -i wasmtime`: present with default features, absent with `--no-default-features`
- [x] `cargo test --bin fledge` (default) — 897 passed, 0 failed
- [x] `cargo clippy --bin fledge -- -D warnings` — clean (also verified with `--no-default-features`)

Closes #445

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,14 @@ sha2 = "0.11.0"
 base64 = "0.22"
 notify = "7"
 shell-words = "1.1.1"
-wasmtime = "46.0.1"
-wasmtime-wasi = "46.0.1"
+wasmtime = { version = "46.0.1", optional = true }
+wasmtime-wasi = { version = "46.0.1", optional = true }
+
+[features]
+default = ["wasm"]
+# Sandboxed WASM plugin runtime (wasmtime + wasmtime-wasi). On by default;
+# disable with `--no-default-features` to drop the wasmtime dependency tree.
+wasm = ["dep:wasmtime", "dep:wasmtime-wasi"]
 
 [build-dependencies]
 toml = "0.8"
--- a/src/plugin/install.rs
+++ b/src/plugin/install.rs
@@ -646,24 +646,36 @@ pub(crate) fn install_plugin(
     }
 
     if manifest.plugin.is_wasm() {
-        for cmd in &manifest.commands {
-            let wasm_path = plugin_dir.join(&cmd.binary);
-            if wasm_path.exists() {
-                println!(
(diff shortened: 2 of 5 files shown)


0xGaspar (approved):
Clean feature-gating. The cfg split is consistent across all three surfaces — the `wasm` module, both run/install dispatch sites, and the protocol host-bridge re-exports (gated to keep the no-default-features build warning-free) — and the disabled path `bail!`s with an actionable message instead of silently degrading. `default = ["wasm"]` keeps current behavior; the dependency-tree win is opt-in via `--no-default-features`, which is the right default. `cargo tree -i wasmtime` present/absent both ways is the proof that matters. LGTM.

PR #466, opened 2026-07-02
Fix: emit specs[] in specsync-delegated spec check --json

### LINKED ISSUE #465: spec check --json emits two different envelopes (built-in vs specsync-delegated)
## Found via dogfooding 1.6.0 on a machine with `specsync` installed

`fledge spec check --json` produces **two incompatible envelope shapes** depending on whether the `specsync` binary is on PATH:

**Built-in path** (no specsync — e.g. CI):
```
{schema_version, action: "spec_check", specs: [...], totals, strict}
```

**Delegated path** (`specsync` installed — e.g. a dev box):
```
{schema_version, action: "spec_check", engine, engine_version, errors, warnings, stale, passed, totals, strict}
```
Note: **no `specs` array** in the delegated shape; it has `errors`/`warnings`/`stale`/`passed`/`engine` instead.

## Impact

1. **Violates the documented `--json` contract.** AGENTS.md documents `spec check --json` as `{…, specs: [...], totals, strict}`. An agent parsing `parsed["specs"]` breaks on any machine with specsync installed.
2. **`tests/spec.rs` fails for any developer with specsync installed.** `cli_spec_check_json_valid` and `cli_spec_check_json_spec_shape` assert `parsed["specs"].is_array()` (spec.rs:154), which is false on the delegated path. Green in CI (no specsync), red locally. (The assert predates recent work — added in #314.)

## Suggested fix

Either:
- **(preferred) Normalize** the specsync-delegated output into fledge's documented envelope (map specsync's result into `{specs: [...], totals, strict}`) so the contract holds regardless of engine, **or**
- Document both shapes in AGENTS.md and make `tests/spec.rs` tolerant of both (assert on `totals`/`strict`, and accept `specs` **or** `errors`/`passed`).

Relates to the test-infra hardening in #447 and spec tooling in #448.

## Evidence
```
$ which specsync
~/.cargo/bin/specsync
$ fledge spec check --json | jq 'keys'
["action","engine","engine_version","errors","passed","schema_version","stale","strict","totals","warnings"]
$ fledge spec check         # content is valid
🔌 Delegating to specsync v4.6.0 (matches CI)
  ✅ 30 specs checked, 0 errors, 0 warnings
```

---
_From hands-on dogfooding of 1.6.0._

0xLeif:
## Summary

`fledge spec check --json` produced **two incompatible envelopes** depending on whether `specsync` is installed:
- **built-in engine** → `{…, engine: "structural", specs: [...], totals, strict}`
- **specsync-delegated** (whenever the `specsync` binary is on PATH — most dev machines) → **no `specs[]` array**

That violated the documented `--json` contract and made `tests/spec.rs`'s `cli_spec_check_json_valid` / `cli_spec_check_json_spec_shape` fail for any developer with specsync installed (green in CI, which has no specsync — so it went unnoticed). Found by dogfooding 1.6.0.

## Fix

specsync's JSON reports only aggregates (a checked count + error/warning/stale lists), never a per-spec breakdown — so the delegated path now populates `specs[]` from **fledge's own structural inventory**, via shared `structural_results` / `spec_result_json` helpers. Both engines now emit an **identical per-spec shape** (`{name, version, status, file_count, section_count, required_count, errors, warnings}`). specsync's aggregate verdict (`passed`, top-level `errors`/`warnings`, `stale`) is preserved and remains authoritative. The built-in path is refactored onto the same helpers — no behavior change, removes a duplicated directory walk.

Helpers are `pub(super)` (module-internal, not exports) so they don't widen the public surface or trip export coverage.

## Test Plan
- [x] `cli_spec_check_json_valid` + `cli_spec_check_json_spec_shape` now pass **with specsync installed** (previously failed locally, passed in CI)
- [x] `fledge lanes run pre-commit` green: fmt + lint + full test suite + spec-check (30 specs, 0 errors, 0 warnings)
- [x] `cargo build --locked` and `cargo build --no-default-features` both clean
- [x] Delegated `spec check --json` output verified: `specs[]` present (len 30), identical keys to the built-in engine
- [x] AGENTS.md + spec Change Log (v13) updated to document the unified envelope

Closes #465


0xLeif:
Good catch — collapsed the two `filter`/`collect` passes in `spec_result_json` into a single pass over `result.issues` in 8d3c5c8. Thanks!

0xGaspar (approved):
Right fix for the contract violation, and it doubles as a real dedup — extracting `structural_results`/`spec_result_json` as `pub(super)` helpers means both engines emit a byte-identical `specs[]` shape with no widened public surface, and the built-in path loses a duplicated directory walk with no behavior change. The layering is sound: specsync stays authoritative for the verdict while `specs[]` comes from fledge's own inventory (which is the only place a per-spec breakdown exists). Good catch that this only ever failed locally because CI has no specsync installed. One documented-but-worth-flagging seam: the delegated `specs[]` is best-effort (empty vec on a config/specs-dir read failure), so a degenerate run could show `specs: []` alongside a nonzero `totals.checked`, and `specs[].errors` is the structural view rather than specsync's export-coverage errors — both are noted in AGENTS.md, so consumers keying off the top-level verdict are fine. LGTM.

PR #468, opened 2026-07-02
Fix: plugins recommend — filter installed plugins and drop duplicates

0xLeif:
## Summary

Two bugs in `fledge plugins recommend`, found by a dogfooding sweep:

1. **Installed plugins were never filtered out.** The filter stripped `fledge-plugin-` from the recommendation repo (`"github"`) and compared it to the registry **name** (`"fledge-plugin-github"`) — never a match. Now it compares the full `owner/repo` against the registry `source` (`"CorvidLabs/fledge-plugin-github"`), which matches exactly.
2. **`fledge-plugin-github` was recommended twice** when a `.github/` directory exists (once from the base list, once from the GitHub-Actions branch). The list is now deduped by repo.

## Test Plan
- [x] `cargo clippy --bin fledge -- -D warnings` / `cargo fmt --check` clean
- [x] New unit test `recommendations_have_no_duplicate_repos` passes (all language arms)
- [x] Behavior in this repo (80 plugins installed): `recommend --json` now filters out installed `fledge-plugin-github`/`deps`/`metrics` and emits **no** duplicate github entry

### DIFF
--- a/src/plugin/recommend.rs
+++ b/src/plugin/recommend.rs
@@ -94,6 +94,11 @@ fn recommendations_for_language(lang: &str) -> Vec<Recommendation> {
         });
     }
 
+    // Dedup by repo: the base list already recommends fledge-plugin-github, and
+    // the `.github` branch above can add it a second time. Keep the first entry.
+    let mut seen = std::collections::HashSet::new();
+    recs.retain(|r| seen.insert(r.repo));
+
     recs
 }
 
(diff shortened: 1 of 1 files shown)


0xLeif:
Applied in 1183c87: both sides now go through `super::normalize_source` (so a full-URL install matches a shorthand repo) and the lookup uses a `HashSet`. `installed_count` now reports the true plugin count. Thanks!

PR #493, opened 2026-07-04
Record attest provenance on merged commits

0xLeif:
From the trust-tooling audit: fledge never ran attest, so its /trust/ attest pip was dark. This adds a dedicated **main-only** `attest` job that runs after the gates (test, lint, audit, spec-check), records a provenance attestation, and pushes `refs/notes/attest`. Public Linux binary, checksum-verified, no token; additive and best-effort, so it can never fail CI.

https://claude.ai/code/session_01ApSS1xj22zGwbUSkRs3mmE

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,3 +189,44 @@ jobs:
           review-on-pr: "true"
           context: ${{ steps.status.outputs.context }}
           job-status: ${{ steps.status.outputs.result }}
+
+  attest:
+    name: Record attestation
+    runs-on: ubuntu-latest
+    needs: [test, lint, audit, spec-check]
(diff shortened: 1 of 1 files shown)


Kyntrin (approved):
Approved. The workflow-only change is narrow, PR checks are green, and the attestation job is correctly scoped to push events on main so the skipped PR check is expected.

Non-blocking hardening notes:
- The binary is verified against a checksum downloaded from the same release. That catches transfer corruption, but not a compromised or replaced release asset. If this job becomes security-critical, pinning an expected SHA in the workflow or using a stronger release provenance check would make the trust boundary clearer.
- The job fetches and then pushes refs/notes/attest once. A concurrent main run could race on that notes ref and log a warning instead of recording provenance. That matches the current best-effort intent, but a retry/re-fetch path would improve recording reliability later.

